PR #8, opened 2025-09-29
Add setup script

granmoe:
Yep, Earth-shattering PR here 😂 But perhaps valuable in the sense of having another person jump into the repo and start to kick the tires.

Just adding a setup script that will take care of cd'ing into payloads/ and installing deps as well as running the existing script to install git hooks. It's possible a contributor might clone the repo and make some changes without setting up the payloads dir and then get an error because ESLint isn't found when the pre-commit hook tries to run it in that dir.

Couple questions/callouts:

* Do you want to use pnpm 8 or pnpm 9+? I can always go back to pnpm 8, but I happened to be using pnpm 9.1.1 locally, thus the updated lockfile version. I have no strong feelings either way.
* Just want to confirm we can go with pnpm and thus delete the extra package-lock.json in payloads/ as I did here.
* I had "format on save" on for md files, which resulted in the Prettier extension inserting some newlines here and there. I figure we'll worry about setting up consistent formatting for root-level files later and add it to the pre-commit hook.


payloads/pnpm-lock.yaml
granmoe, line 1: Sorry, I was on pnpm 9+, thus the updated lockfile version. We just need to pick between 8 or 9+

scripts/setup.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Development environment setup script for LLMIR
+# Run this after cloning the repository
+
+set -e
+
+echo "🚀 Setting up LLMIR for development..."
+
+# Check if we're in the right directory
+if [ ! -f "Cargo.toml" ]; then
```
granmoe: This isn't too resilient of a check. Thought of checking for the existence of scripts/install-hooks.sh, but that's not bulletproof either.

I suppose we could take `__filename`, calculate the absolute path to the repo based on that, and check that cwd equals that. I'll make this update tonight.

payloads/pnpm-lock.yaml
ankrgyl: ah, apparently you can specify 

```
"packageManager": "pnpm@9.12.0"
```

in package.json and use it. I don't care which version it is (later is likely better). We just use an old version for our main braintrust repo

scripts/setup.sh
ankrgyl: seems fine

ankrgyl (approved):

src/providers/openai/convert.rs
```diff
@@ -692,7 +692,7 @@ impl TryFromLLM<openai::OutputItem> for openai::InputItem {
             .map(|mr| match mr {
                 openai::MessageRole::Assistant => openai::InputItemRole::Assistant,
             })
-            .or_else(|| {
+            .or({
```
granmoe, line 695: Little lint error caught by clippy

PR #18, opened 2025-10-10
Add Go bindings

remh:
This PR adds go bindings for linga.

The go API currently supports passing payloads through go []maps.
Those payloads are then marshalled to json to be passed to the rust code. 
It might make sense at some point to support an API that would accept json payloads directly to avoid uncessary serialization / deserialization.


.github/workflows/ci.yml
```diff
@@ -153,3 +153,49 @@ jobs:
...
+      run: |
+        export LD_LIBRARY_PATH="${{ github.workspace }}/target/release:$LD_LIBRARY_PATH"
+        golangci-lint config verify
+        golangci-lint run
+
+    - name: Run Go tests
+      working-directory: bindings/golang
+      run: |
+        export LD_LIBRARY_PATH="${{ github.workspace }}/target/release:$LD_LIBRARY_PATH"
+        go test -v
```
clutchski: nitpick: I'd prefer all of this to live in a makefile and github actions only calls "make ci", then it's reproducible locally,

bindings/golang/lingua_test.go
```diff
@@ -0,0 +1,238 @@
...
+	linguaMsgs, err := AnthropicMessagesToLingua(anthropicMsgs)
+	require.NoError(t, err)
+	require.Len(t, linguaMsgs, 1)
+	assert.Equal(t, "user", linguaMsgs[0]["role"])
+
+	// Test converting back to Anthropic format
+	backToAnthropic, err := LinguaToAnthropicMessages(linguaMsgs)
+	require.NoError(t, err)
+	require.Len(t, backToAnthropic, 1)
+	assert.Equal(t, "user", backToAnthropic[0]["role"])
```
clutchski: any reason we can't assert the whole messages is equal?

.github/workflows/ci.yml
remh: fixed

bindings/golang/lingua_test.go
remh: fixed

PR #22, opened 2025-10-20
typescript: browser wasm exports

choochootrain:


bindings/typescript/tsup.config.ts
```diff
@@ -0,0 +1,41 @@
+import { defineConfig } from "tsup";
+import { readFileSync, writeFileSync } from "fs";
+import { join } from "path";
+
+export default defineConfig([
+  {
+    entry: {
+      index: "src/index.ts",
+    },
+    format: ["esm"],
```
ibolmo, line 10: curious if we'll get cjs requests

ibolmo (approved):
LGTM. I do wonder about cjs support, but we can get there if we need to get there.

PR #23, opened 2025-10-20
spike: adding ai-sdk to lingua

ibolmo:


ibolmo:
this is a good start, but closing. fairly easy to add more cases and discovered a few surprises if we were to add multiple providers that are not specifically ai providers. captured that internally, and we'll come back to supporting more ai tools in the future.

PR #24, opened 2025-10-21
try explicit ai-sdk entrypoints

ibolmo:


ibolmo:
this could work but going to punt on this for now

PR #27, opened 2025-10-30
Add turbo repo config for supporting caching

KevinGreen:
Without this, turbo build with remote caching doesn't properly capture all the build objects so if there's a cache hit, tests fail to find the wasm file.

### DIFF
--- a/bindings/typescript/turbo.json
+++ b/bindings/typescript/turbo.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "outputs": [
+        "dist/**",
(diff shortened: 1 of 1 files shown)


bindings/typescript/turbo.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
```
choochootrain, line 3: oh weird i thought this extended from the root in the repo

choochootrain (approved):

bindings/typescript/turbo.json
KevinGreen: Maybe because this package is build as a submodule it finds the turbo.json in the root of braintrust

PR #26, opened 2025-10-24
Default to returning generated provider types for conversion functions but support accepting a generic

granmoe:
Noticed while writing an example against the TS bindings that helpers that convert _from_ lingua return `unknown[]`.

With these changes, we can now pass a generic for a structurally compatible type:

Screen Shot 2025-10-24 at 11 08 40
Screen Shot 2025-10-24 at 11 08 54

And if we pass no generic, we get back Lingua's generated type for the target format:

Screen Shot 2025-10-24 at 11 09 18

---

Before, we couldn't return the target provider types here because we didn't export them to TS. Now, we mark them for export in `generate-types.rs` since they're part of our public API surface, and we ensure that any transitive dependencies that are pulled in end up in the same provider directory with their corresponding "entry point" types by using the `export_to` ts-rs attribute.

With the types exported, we can use them in converters, and allow callers to pass their own generic to get back any type they want.

Just throwing this PR up as one potential way to solve this problem.

Note that there's a ton of generated code in bindings/typescript/src/generated/{provider}, and that makes up the bulk of the diff. I'll call out the meat of the changes with comments.

PR #28, opened 2025-11-05
typescript: integration tests for node and browser exports

choochootrain:
Add some basic tests to verify the typescript build artifacts for node and browser


choochootrain:
@granmoe this is a first pass on what we discussed in #26. i didn't write a test to verify in the jsdom environment here but vitest lets you do this pretty easily if we find it useful

granmoe (approved):
Awesome, LGTM! 🙌

PR #29, opened 2025-11-05
Responses file/reasoning fixes and more test cases

choochootrain:
Defined a few more cases we might want to care about for multi-turn, reasoning, and multi-modal for all providers and fix the discovered issues:

responses:
* Fix file and reasoning part handling
* Modify responses tests to roundrip `OutputItem -> Message -> OutputItem` instead of going through `InputItem` - this is consistent with other provider tests, we can reconsider if necessary

chat completions:
* Fix file part handling

test improvements:
* Use colorized diffs for all test runner failure messages
* Removed duplicate toolCallRequest case from `advanced.ts`
* Moved `reasoningWithOutput` to `simple.ts`


knjiang (approved):
seems good to me.

PR #30, opened 2025-11-12
Use wrapper crate for `serde_json` with `arbitrary_precision` enabled

knjiang:
see main repo PR ->  https://github.com/braintrustdata/braintrust/pull/8779


scripts/generate-types.rs
```diff
@@ -1088,13 +1106,26 @@ fn add_ts_type_annotations(content: &str) -> String {
             let has_ts_attr = prev_line.starts_with("#[ts(");
 
             // Determine if we need to add ts annotation
-            // Check the FULL line for serde_json::Value (handles complex generic types)
-            let needs_ts_annotation = line.contains("serde_json::Value");
+            // Check the FULL line for serde_json::Value or serde_json::Value (handles complex generic types)
+            let has_serde_json_value =
+                line.contains("serde_json::Value") || line.contains("serde_json::Value");
```
ankrgyl: this codeo looks supicious. why did we have to change this file at all?

knjiang: yep, i think we just need to add a hook to ensure `use crate::serde-json` imports top-level. 

i've fixed it, it seems like the only unexpected changes is new updates to the google protobuf stuff. it seems we always download latest so that coudl be expected?

ankrgyl (approved):

PR #31, opened 2025-11-17
Gateway transformations

remh:
## Summary

Add format detection, request transformation infrastructure, and OpenAI-specific preprocessing for gateway integration.

### Changes

- **Format Detection**: Struct-based validation to detect incoming request formats (OpenAI, Anthropic, Google, Bedrock, Mistral) via `is_valid_for_format()` and `validate_or_transform()` APIs
- **Request Transformation Pipeline**: Convert payloads between any supported format through universal message representation with automatic provider defaults (e.g., `max_tokens` for Anthropic)
- **OpenAI Request Transformations**: Request preprocessing via `OpenAIRequestTransformer` builder - reasoning model handling, structured output fallback, provider-specific field sanitization
- **Universal Message Transforms**: System message extraction and consecutive message flattening for providers that require them
- **Media Utilities**: URL parsing and base64 conversion utilities for images/PDFs

### New Public API
```rust
use lingua::{ProviderFormat, TransformResult, TransformError};

// Check if payload is valid for target format (struct-based validation)
if lingua::is_valid_for_format(&payload, ProviderFormat::Anthropic) {
    // Use payload as-is (pass-through)
}

// Full request transformation with provider defaults
let result = lingua::transform_request(&payload, ProviderFormat::Anthropic)?;
match result {
    TransformResult::PassThrough => { /* use original payload */ }
    TransformResult::Transformed { payload, source_format } => {
        // payload has been transformed and defaults applied
    }
}

// Convert to/from universal message format
let messages: Vec<Message> = lingua::to_universal(&payload, ProviderFormat::OpenAI)?;
let anthropic_messages = lingua::from_universal(&messages, ProviderFormat::Anthropic)?;

// Apply provider-specific defaults (e.g., max_tokens for Anthropic)
lingua::apply_provider_defaults(&mut payload, ProviderFormat::Anthropic);#### OpenAI-specific transformations

use lingua::providers::openai::transformations::{OpenAIRequestTransformer, TargetProvider};

// Transform OpenAI requests for downstream providers
OpenAIRequestTransformer::new(&mut request)
    .with_target_provider(TargetProvider::Mistral)
    .with_provider_metadata(metadata)
    .with_supports_streaming(Some(true))
    .transform()?;#### Universal message transforms

use lingua::universal::{extract_system_messages, flatten_consecutive_messages};

// Extract system messages (for Anthropic, Google, Bedrock)
let system_prompts = extract_system_messages(&mut messages);

// Merge consecutive same-role messages
flatten_consecutive_messages(&mut messages);### Supported Formats
```
| Format | Detection | To Universal | From Universal |
|--------|-----------|--------------|----------------|
| OpenAI | ✓ | ✓ | ✓ |
| Anthropic | ✓ | ✓ | ✓ |
| Google | ✓ | ✓ | ✓ |
| Bedrock (Converse) | ✓ | ✓ | ✓ |
| Mistral | ✓ | ✓ (via OpenAI) | ✓ (via OpenAI) |

### Testing

- Payload-based tests in `payloads/transformations/` covering provider sanitization, reasoning models, structured output, and cross-format conversion
- Unit tests for all transformation functions


docs/openai-transformations.md
```diff
@@ -0,0 +1,35 @@
+<!-- Tracking document for OpenAI payload transformations migrated from proxy.ts -->
```
ankrgyl: is this useful to keep around for LLMs?

src/providers/openai/transformations.rs
```diff
@@ -0,0 +1,393 @@
+/*!
```
ankrgyl, line 1: I think a better way to model these transformations is using a set of "Capabilities" that can be configured per model provider, model, etc. and get passed into the transformation logic when converting lingua -> a specific provider (or just operate lingua -> lingua)

docs/openai-transformations.md
remh: it's both for me and LLMs to keep until completion

src/providers/openai/transformations.rs
remh: not super familiar with Rust Capabilities, looking into it now

ankrgyl (approved):

src/processing/catalog.rs
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+use crate::capabilities::ProviderFormat;
+use std::sync::OnceLock;
+
+/// Type alias for the catalog lookup function.
+type CatalogLookupFn = Box<dyn Fn(&str) -> Option<ProviderFormat> + Send + Sync>;
+
+/// Global catalog lookup function, injected at startup.
+static CATALOG_LOOKUP: OnceLock<CatalogLookupFn> = OnceLock::new();
```
choochootrain: maybe out of scope since this PR is already getting big but we probably want a way to periodically refresh the catalog like we do in the proxy: braintrustdata/braintrust-proxy#279

src/universal/transform.rs
```diff
@@ -0,0 +1,391 @@
+/*!
+Lingua → Lingua transformations that run before converting to a provider format.
+
+This module provides universal transformations that are common across multiple
+providers, mirroring the proxy's behavior:
+
+- **Message flattening**: Merge consecutive messages of the same role
```
choochootrain, line 7: would we ever need to reverse this transformation in responses? if you are going through the gateway using an openai-style request you might expect the message indices in the response to match the message indices in your request (pre-flattening)

i dont think it's super important right now but this might end up being a general problem where we need to track some state on transformed gateway requests so we can build a proper response in the user's expected format

choochootrain (commented):
looks reasonable at a high level. i had 2 questions that can be addressed outside this PR, just thinking out loud about what end to end behavior we want

src/processing/catalog.rs
remh: thanks, i think this makes sense. I'll add it to my personal todo.

src/universal/transform.rs
remh: hm good question, @ankrgyl what do you think?

src/capabilities/format.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    /// Returns true if this format is a known, supported format.
+    pub fn is_known(&self) -> bool {
+        !matches!(self, ProviderFormat::Unknown)
+    }
+}
+
+impl std::fmt::Display for ProviderFormat {
```
ankrgyl, line 42: nit: just implement the body of `as_str` here (assuming you are not calling `.as_str()` anywhere directly)

src/capabilities/format.rs
```diff
@@ -0,0 +1,98 @@
...
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "openai" => Ok(ProviderFormat::OpenAI),
+            "anthropic" => Ok(ProviderFormat::Anthropic),
+            "google" => Ok(ProviderFormat::Google),
+            "mistral" => Ok(ProviderFormat::Mistral),
+            "converse" | "bedrock" => Ok(ProviderFormat::Converse),
+            _ => Ok(ProviderFormat::Unknown),
```
ankrgyl: i think it would be slightly better for this to return an error and then let the caller coerce that to unknown (or add a method called `try_from_str()` that does that).

src/lib.rs
```diff
@@ -10,6 +10,18 @@ pub mod universal;
 pub mod util;
 pub mod validation;
 
+pub use providers::openai::transformations;
```
ankrgyl: i'd rather let the caller import specific paths than re-export these provider-specific paths

src/processing/detect.rs
```diff
@@ -0,0 +1,764 @@
...
+
+    // 2. Fall back to content-based heuristics
+    detect_from_content(payload, model, strict)
+}
+
+/// Detect format from a JSON string.
+///
+/// Internal convenience wrapper that parses the JSON first.
+#[cfg(test)]
+fn detect_format_from_str(payload: &str, strict: bool) -> Result<DetectedPayload, DetectionError> {
```
ankrgyl: do we really need this function? Can the caller just do this?

src/processing/catalog.rs
ankrgyl: I think if you change this to be an arc swap it should make that easy (each time you call `set_catalog_lookup` it'll fix it)

src/providers/anthropic/detect.rs
```diff
@@ -0,0 +1,441 @@
...
+///
+/// let anthropic_payload = r#"{
+///     "model": "claude-3-5-sonnet-20241022",
+///     "messages": [{"role": "user", "content": "Hello"}],
+///     "max_tokens": 1024
+/// }"#;
+///
+/// assert!(is_anthropic_format(anthropic_payload).unwrap());
+/// ```
+pub fn is_anthropic_format(payload: &str) -> Result<bool, DetectionError> {
```
ankrgyl: I'm getting a bit confused about the intended interface. Is the idea that each format (eg Anthropic) implements a format detector, which accepts `Value`? Or do we think that deserializing directly into the struct for each provider from a `str` is faster? In either case, I think we shuold settle on exactly one input type (either `Value` or `str`).

src/providers/bedrock/detect.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::serde_json::json;
+
+    #[test]
+    fn test_bedrock_converse_with_model_id() {
+        let payload = json!({
```
ankrgyl, line 199: the problem with these tests is that we don't really assert that these are valid bedrock objects. I think it would be better to populate the actual bedrock struct, serialize it to `Value`, and then call `is_bedrock_converse`. That way, if we update the provider types and no longer have valid Bedrock structs, these tests will fail.

ankrgyl: (and ofc same for every provider)

ankrgyl (changes requested):
just for bookkeeping purposes, undoing my quick approval

src/capabilities/format.rs
remh: addressed in https://github.com/braintrustdata/lingua/pull/31/commits/afc3d3377b3d307d16f7938148495b8e396e931d

src/capabilities/format.rs
remh: addressed in https://github.com/braintrustdata/lingua/pull/31/commits/2c4d25bad6cb890ef3392146ed43e8e435cdbbf7

src/lib.rs
remh: fixed

src/providers/bedrock/detect.rs
remh: addressed

src/providers/anthropic/detect.rs
remh: fixed in https://github.com/braintrustdata/lingua/pull/31/changes/4862fb5c4aeaac269e3a223a29512da23cfa9295

src/processing/catalog.rs
remh: Fixed in https://github.com/braintrustdata/lingua/pull/31/commits/8046ef2ed8146bef8fee1cb2579f04a70deca3bc

src/universal/transform.rs
remh: will think about this later

knjiang:
* **#31** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/31?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#38](https://github.com/braintrustdata/lingua/pull/38) <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/38?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#40](https://github.com/braintrustdata/lingua/pull/40) <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/40?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/31?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

remh:
Superseeded by https://github.com/braintrustdata/lingua/pull/50

PR #33, opened 2025-11-30
Universal tools (part 1)

granmoe:
This is the first of several PRs needed to support typed tool definitions and Lingua conversion for them. This PR includes only the updates to `generate-types.rs` to generate fully-typed tool definition types, as well as the resulting generated types. Converters, tests, examples, etc will come in subsequent PRs.

A template-based type generation approach produces discriminated union enums for tools scraped from OpenAPI specs. Instead of relying on quicktype's flat struct output (which combines all tool fields into a single struct with many optionals), we now generate proper Rust enums with `#[serde(tag = "type")]` for type-safe tool handling.

We also scrape fully-typed provider tools directly out of their specs via helper methods found in the new, inline `tool_generator` module in `generate-types.rs`. We're able to generate individual structs for each tool type (CustomTool, BashTool20250124, WebSearchTool, etc.), and then generate a `Tool` enum with tagged serde deserialization.

Note this currently is only implemented for Anthropic and OpenAI - Google to be added later.

## Why direct / template-based generation?

Quicktype generates **union structs** instead of discriminated enums. For tools, this produces:

```rust
// What quicktype generates - a flat struct with all fields
pub struct Tool {
    pub name: String,
    pub tool_type: Option<ToolType>,
    // Client tool fields
    pub description: Option<String>,
    pub input_schema: Option<InputSchema>,
    // Web search fields
    pub allowed_domains: Option<Vec<String>>,
    pub blocked_domains: Option<Vec<String>>,
    pub user_location: Option<UserLocation>,
    // Text editor fields
    pub max_characters: Option<i64>,
    // Computer use fields
    pub display_width: Option<i64>,
    pub display_height: Option<i64>,
    // ... 20+ more fields, most irrelevant for any given tool
}
```

Compare to what we generate with this PR:

```rust
// Proper discriminated union
#[serde(tag = "type")]
pub enum Tool {
    #[serde(rename = "bash_20250124")]
    Bash20250124(BashTool20250124),

    #[serde(rename = "web_search_20250305")]
    WebSearch20250305(WebSearchTool20250305),

    #[serde(untagged)] // If no type passed, fallback to this (eg for client-defined tools)
    Custom(CustomTool),
}

// struct for each tool
pub struct BashTool20250124 {
    pub name: String,
    pub cache_control: Option<serde_json::Value>,
}

pub struct WebSearchTool20250305 {
    pub name: String,
    pub allowed_domains: Option<Vec<String>>,
    pub blocked_domains: Option<Vec<String>>,
    pub max_uses: Option<i64>,
    // ... only fields relevant to web search
}
```

TLDR - I tried really hard to get quicktype to work or a typify -> quicktype pipeline, but you basically end up needing to do a ton of post-processing and fight with what these libraries want to do (all the merging etc).

### scripts/generate-types.rs

Added two new modules:

**`tool_generator`** - Extracts and generates tool types:
- `extract_tool_schemas()` - Finds tool definitions in OpenAPI specs
- `generate_tool_structs()` - Creates individual tool struct code
- `generate_tool_enum()` - Creates the discriminated `Tool` enum
- `replace_tool_struct_with_enum()` - Injects generated code into quicktype output

**`schema_converter`** - JSON Schema to Rust type conversion:
- `schema_type_to_rust()` - Maps JSON Schema types to Rust equivalents
- `to_rust_field_name()` - Converts property names to valid Rust identifiers

### Serde deserialization strategy

The `Tool` enum uses a hybrid tagging approach:

```rust
#[serde(tag = "type")]
pub enum Tool {
    // Provider tools use explicit tag matching
    #[serde(rename = "bash_20250124")]
    Bash20250124(BashTool20250124),

    // Client tools use untagged fallback (no "type" field required)
    #[serde(untagged)]
    Custom(CustomTool),
}
```

This allows:
1. Provider tools to match by their `type` field value
2. Client tools (like Anthropic's custom tools) to deserialize without a `type` field

## What's NOT in this PR

This PR focuses only on type generation. The following will come in separate PRs:
- Universal tool types (`src/universal/tool.rs`)
- Tool conversion logic (`src/providers/*/convert.rs`)
- WASM exports for tool conversion
- TypeScript tool conversion functions
- Tests

## Open questions

- How fancy should we get in building the Rust types for provider tools input schemas? E.g. some things that used to be string unions now are just strings. We could build this out if desired.
- How to handle unknown provider tool `type` values? I talked to Opus 4.5 about this:

<details>

<summary>Conversation with Opus 4.5 about options for handling unknown tool types</summary>

Currently there's no built-in way to pass through an arbitrary provider tool definition.

  The limitation: The ProviderTool struct has tool_type, name, and config fields, but the converter explicitly matches on known
  tool_type values. If a new tool type appears, the conversion fails.

  Potential solutions (not implemented):

  1. A "raw" or "passthrough" tool type - A special tool_type value (e.g., "raw") that tells the converter to serialize the config field
   directly as the tool definition:
  Tool::Provider(ProviderTool {
      tool_type: "raw".to_string(),
      name: None,
      config: Some(serde_json::json!({
          "type": "new_fancy_tool",
          "some_field": "value"
      })),
  })
  2. A separate RawTool variant - Add a new variant to the Tool enum specifically for passthrough:
  Tool::Raw(serde_json::Value)
  3. Soft-fail with warning - Instead of erroring, convert unknown types to a raw JSON representation that gets passed through, possibly
   with a warning logged.

  The tradeoff is between safety (catching typos/mistakes in tool types) vs flexibility (forward compatibility). The current strict
  approach catches errors early but requires Lingua updates for new provider features.

## Current state

  The conversion logic explicitly matches on known tool_type strings:

  - OpenAI (src/providers/openai/convert.rs:1529-1663): Matches computer_use_preview, code_interpreter, web_search, file_search, mcp,
  image_generation, local_shell, web_search_preview
  - Anthropic (src/providers/anthropic/convert.rs:766-849): Matches bash_20250124, text_editor_*, web_search_20250305

  Unknown types return errors like:
  Err(ConvertError::UnsupportedInputType {
      type_info: format!("Unsupported OpenAI provider tool type: '{}'...", unknown)
  })

  ---
  Analysis of approaches

  Approach 1: "raw" or "passthrough" tool type

  Implementation complexity: Low

  Add a special case in each provider's conversion:
  "raw" | "passthrough" => {
      // Pass config directly, wrapping with provider's expected structure
      let config = provider_tool.config.unwrap_or_else(|| serde_json::json!({}));
      // Provider-specific wrapping logic here
  }

  Pros:
  - Minimal code change (one match arm per provider)
  - Explicit opt-in - users know they're bypassing type safety
  - Uses existing config field

  Cons:
  - Magic string feels inconsistent with the strongly-typed design
  - The config field's structure may not match what providers expect. OpenAI tools have {"type": "...", ...} at root level, but config
  is nested inside ProviderTool
  - Unclear semantics - does "raw" mean the same thing for OpenAI vs Anthropic?
  - Harder to implement bidirectional conversion - how would you detect a raw tool when converting from provider format?

  Alignment with Lingua goals: Moderate. Maintains explicit control but introduces a non-type-safe escape hatch.

  ---
  Approach 2: Separate Tool::Raw(serde_json::Value) variant

  Implementation complexity: Medium

  pub enum Tool {
      Client(ClientTool),
      Provider(ProviderTool),
      Raw(serde_json::Value),  // New variant
  }

  Then in converters:
  Tool::Raw(value) => {
      // Pass through directly, provider decides how to handle
      serde_json::from_value(value)?
  }

  Pros:
  - Clean, self-documenting API
  - Type-safe at the enum level
  - Flexible - any valid JSON structure is accepted
  - Consistent across providers
  - Clear intent - users explicitly choose raw mode

  Cons:
  - Breaking change to the core Tool enum
  - TypeScript bindings need updating (Tool.ts)
  - Loses semantic information - can't introspect what kind of tool it is
  - Bidirectional conversion is ambiguous - when converting from provider format, how do you decide if something should become Tool::Raw
   vs Tool::Provider?
  - May encourage users to bypass the type system unnecessarily

  Alignment with Lingua goals: Mixed. Explicit but sacrifices type safety for flexibility.

  ---
  Approach 3: Soft-fail with warning

  Implementation complexity: Medium-High

  unknown => {
      // Instead of error, pass through with warning
      log::warn!("Unknown tool type '{}', passing through as raw", unknown);
      // Construct provider-native format from config
      serde_json::from_value(provider_tool.config.unwrap_or_default())?
  }

  Pros:
  - Most forward-compatible - new provider tools work immediately
  - No user action required when new tools appear
  - Graceful degradation

  Cons:
  - Directly violates Lingua's "explicit error handling" principle - errors become silent
  - Typos in tool_type become runtime bugs instead of compile/conversion-time errors
  - No logging infrastructure exists in Lingua (it's a library, not an application)
  - Makes debugging much harder - "why isn't my tool working?" becomes mysterious
  - Loses the safety net that catches mistakes early

  Alignment with Lingua goals: Poor. Goes against core principle of explicit error handling.

  ---
  Fourth option worth considering: Provider-level fallback variant

  Instead of modifying Tool, add a Fallback variant to the provider-specific enums:

  // In openai/generated.rs or a wrapper
  pub enum Tool {
      Function(FunctionTool),
      CodeInterpreter(CodeInterpreterTool),
      // ... existing variants
      Fallback(serde_json::Value),  // Catches unknown types
  }

  This handles unknown types at deserialization time, keeping the universal format clean.

  Pros:
  - Handles unknown types during parsing gracefully
  - Universal Tool enum stays clean and type-safe
  - Provider-specific handling of unknowns
  - Bidirectional conversion is cleaner - Fallback maps to Tool::Raw or a specific ProviderTool

  Cons:
  - Requires modifying generated types (or post-processing them)
  - Different providers may need different fallback strategies
  - Still loses type information for unknown tools

  ---
  Recommendation

  Given Lingua's goals of type safety and explicit error handling, I'd lean toward Approach 2 (Tool::Raw variant) with a refinement:

  pub enum Tool {
      Client(ClientTool),
      Provider(ProviderTool),
      /// Raw provider-specific tool definition (escape hatch)
      /// 
      /// Use this when Lingua doesn't yet support a new provider tool type.
      /// The value is passed directly to the provider without transformation.
      Raw {
          /// Target provider (e.g., "openai", "anthropic")
          provider: String,
          /// Raw tool definition in provider's expected format
          definition: serde_json::Value,
      },
  }

  Why this variant is better:
  1. Explicit - users consciously choose to bypass type safety
  2. Provider-aware - converter knows which provider to target
  3. Self-documenting - the struct fields explain what's needed
  4. Forward-compatible - new tools work immediately
  5. Doesn't hide errors - using Tool::Provider with wrong type still errors

  The tradeoff is that it's more invasive than Approach 1, but it's cleaner and more aligned with Lingua's design philosophy than either
   a magic string or silent pass-through.

  Want me to explore the implementation details of any of these approaches further?

> How would that option look on the calling side? 

⏺ Here's how the Tool::Raw variant would look from the caller's perspective:

  Current approach (requires Lingua update for new tools)

  // TypeScript
  const tools: Tool[] = [
    {
      type: "provider",
      tool_type: "web_search_20250305",  // Must be a known type
      name: "search",
      config: { max_uses: 5 }
    }
  ];

  // If you try an unknown type, conversion fails:
  const tools: Tool[] = [
    {
      type: "provider",
      tool_type: "new_fancy_tool_20250601",  // ERROR: Unsupported tool type
      config: { some_field: "value" }
    }
  ];

  With Tool::Raw (Approach 2 refined)

  // TypeScript
  const tools: Tool[] = [
    // Known tools still work the same way
    {
      type: "provider",
      tool_type: "web_search_20250305",
      config: { max_uses: 5 }
    },

    // New/unknown tools use the raw escape hatch
    {
      type: "raw",
      provider: "anthropic",
      definition: {
        // Exact structure Anthropic expects - you're responsible for correctness
        type: "new_fancy_tool_20250601",
        name: "fancy",
        some_field: "value"
      }
    }
  ];

  // Rust
  let tools: Vec<Tool> = vec![
      // Known tool - type-safe
      Tool::Provider(ProviderTool {
          tool_type: "web_search_20250305".to_string(),
          name: Some("search".to_string()),
          config: Some(serde_json::json!({ "max_uses": 5 })),
      }),

      // Unknown tool - raw passthrough
      Tool::Raw {
          provider: "anthropic".to_string(),
          definition: serde_json::json!({
              "type": "new_fancy_tool_20250601",
              "name": "fancy",
              "some_field": "value"
          }),
      },
  ];

  Comparison with Approach 1 ("raw" magic string)

  // Approach 1 - magic string
  {
    type: "provider",
    tool_type: "raw",  // Magic value
    config: {
      // What goes here? The whole tool? Just the config?
      // Unclear and provider-dependent
    }
  }

  // Approach 2 - explicit variant
  {
    type: "raw",
    provider: "anthropic",
    definition: { /* exactly what the provider expects */ }
  }

  Real-world example: Hypothetical new Anthropic tool

  Say Anthropic releases "code_sandbox_20250601" tomorrow:

  // With Tool::Raw, works immediately:
  const tools: Tool[] = [
    {
      type: "raw",
      provider: "anthropic",
      definition: {
        type: "code_sandbox_20250601",
        name: "sandbox",
        language: "python",
        timeout_ms: 30000
      }
    }
  ];

  // Convert to Anthropic format
  const anthropicRequest = lingua.toAnthropic({ tools, messages });
  // The raw definition is passed through directly into the tools array

  Error behavior comparison

  // Typo in known tool type - GOOD: fails fast with clear error
  {
    type: "provider",
    tool_type: "web_serach_20250305",  // typo
    // Error: Unsupported Anthropic provider tool type: 'web_serach_20250305'
  }

  // Raw tool - user is responsible, passes through
  {
    type: "raw",
    provider: "anthropic",
    definition: { type: "web_serach_20250305" }  // typo passes through
    // No Lingua error - Anthropic API will reject it
  }

  The key UX difference

  | Scenario               | Tool::Provider                      | Tool::Raw               |
  |------------------------|-------------------------------------|-------------------------|
  | Known tool type        | Type-safe, guided                   | Works but unnecessary   |
  | Unknown tool type      | Error (must wait for Lingua update) | Works immediately       |
  | Typo in tool type      | Caught at conversion time           | Caught at API call time |
  | Config structure wrong | May be caught                       | Not caught by Lingua    |

  The Tool::Raw variant is explicitly an escape hatch - you trade Lingua's safety net for immediate forward compatibility. The separate
  variant makes this tradeoff visible in the code.
</details>


scripts/generate_types/main.rs
```diff
@@ -184,7 +186,6 @@ fn generate_openai_types_with_quicktype(
         .arg("--derive-debug")
         .arg("--derive-clone")
```
granmoe, line 180: Is this supported in a certain quicktype version? I'm on the 23.2.6 npm package, which is latest, and this was throwing as an unsupported option

scripts/generate-types.rs
```diff
@@ -1488,3 +1517,628 @@ fn fix_google_type_references(content: String) -> String {
...
+// Tool Generator Module
+// =============================================================================
+//
+// This module contains utilities for generating tool types (structs and enums)
+// from OpenAPI schemas. Instead of relying on quicktype for tool generation
+// (which requires extensive post-processing), we generate tool structs directly
+// from schema analysis. Tool schemas are simple flat objects with primitive
+// fields, making direct codegen cleaner.
+
+mod tool_generator {
```
granmoe: Threw this all in an inline module to keep things somewhat orthogonal and organized. (I love Rust's module system...wish it existed in JS land!)

scripts/generate_types/main.rs
ankrgyl: Here's what i have:

```
$ quicktype --version
quicktype version 23.2.6
Visit quicktype.io for more info.
```

scripts/generate-types.rs
ankrgyl: even better would be to put this in a new file :)

scripts/generate-types.rs
```diff
@@ -1488,3 +1517,628 @@ fn fix_google_type_references(content: String) -> String {
...
+    fn get_schemas(
+        spec: &serde_json::Value,
+    ) -> Option<&serde_json::Map<String, serde_json::Value>> {
+        spec.get("components")
+            .and_then(|c| c.get("schemas"))
+            .and_then(|s| s.as_object())
+    }
+
+    /// Find the position after the closing brace that matches the opening brace at `open_pos`
+    fn find_closing_brace(content: &str, open_pos: usize) -> Option<usize> {
```
ankrgyl: does this parse the generated rust code, and then convert it? If so, would it be more straightforward to use the JSON schema and generate the correct Rust code directly?

scripts/generate-types.rs
granmoe: Done! Thanks for the nudge - was a bit wary to do this as a Rust n00b, but it ended up being pretty straightforward and is much much nicer 🙂

scripts/generate_types/main.rs
granmoe: @ankrgyl Are you possibly using a different version in lingua somehow? This option just fails for me on 23.2.6:

```
$ quicktype -v
quicktype version 23.2.6
Visit quicktype.io for more info.

$ echo '{"type": "object", "properties": {"name": {"type": "string"}}}' | quicktype --src-lang schema --lang rust --derive-default 2>&1


Error: Option parsing failed: .
```

whereas without it, I get:

```
$ echo '{"type": "object", "properties": {"name": {"type": "string"}}}' | quicktype --src-lang schema --lang rust 2>&1                 


// Example code that deserializes and serializes the model.
// extern crate serde;
// #[macro_use]
// extern crate serde_derive;
// extern crate serde_json;
//
// use generated_module::TopLevel;
//
// fn main() {
//     let json = r#"{"answer": 42}"#;
//     let model: TopLevel = serde_json::from_str(&json).unwrap();
// }

use serde::{Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
pub struct TopLevel {
    name: Option<String>,
}
```

scripts/generate-types.rs
granmoe: This is only used to remove the `pub struct Tool` that quicktype generates. We could try to get quicktype to not generate this, but it would be hard since a bunch of schemas reference it. I think beating quicktype into submission would end up being more complex and brittle than the isolated bit of post-processing / string surgery here.

Makefile
```diff
@@ -7,6 +7,16 @@ help: ## Show this help message
 	@echo ""
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2}'
 
+generate-provider-types: ## Regenerate provider types from OpenAPI specs (usage: make generate-provider-types PROVIDER=openai)
```
ankrgyl, line 10: we should add a top level rule called `generate-all-providers` that calls `./pipelines/generate-provider-types.sh`. Or, just replace this with that rule (unless you really want to call individual providers separately).

src/providers/anthropic/generated.rs
```diff
@@ -784,49 +784,129 @@ pub enum ToolChoiceType {
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, TS)]
 #[ts(export_to = "anthropic/")]
-pub struct Tool {
+pub struct CustomTool {
     /// Create a cache control breakpoint at this content block.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub cache_control: Option<CacheControlEphemeral>,
+    #[ts(type = "any")]
```
ankrgyl: this likely predates this change but i'd much rhater use `unknown` than `any`

ankrgyl (commented):
Looking great, just left a couple nit comments

ankrgyl (commented):
If this works for writing typescript code now, could we add some tests or examples?

granmoe:
> If this works for writing typescript code now, could we add some tests or examples?

For sure, I've got those written out in the main feature branch, but I just wanted to keep this PR limited to only the types and type gen code for ease of review. Cool if I add tests and examples in one of the following PRs?

ankrgyl:
> > If this works for writing typescript code now, could we add some tests or examples?
> 
> For sure, I've got those written out in the main feature branch, but I just wanted to keep this PR limited to only the types and type gen code for ease of review. Cool if I add tests and examples in one of the following PRs?

Yes!

ankrgyl (approved):

Makefile
granmoe: Done. I did find running individual providers to be useful for iterating on `generate-types`, hopefully we'll get more utility out of it as times goes on, but I'm also cool deleting it and adding back later if needed

src/providers/anthropic/generated.rs
granmoe: Same - thank you for catching! Let me track down where this is coming from and fix it (prob just the hardcoded thing in `add_ts_type_annotations`)

PR #47, opened 2025-12-29
pnpm 10

ibolmo:


ankrgyl (commented):
can we avoid making mise a requirement for this repo? i'd rather specify the required pnpm version in package.json

ibolmo:
> can we avoid making mise a requirement for this repo? i'd rather specify the required pnpm version in package.json

sounds good. i'll send up a fix in a bit.

ankrgyl (approved):

PR #59, opened 2026-01-14
add anthropic messages parameter test cases

knjiang:
Add Anthropic Messages API parameter test cases to the existing `params.ts` file for cross-provider mapping coverage.

  New test coverage, got the request body from https://docs.anthropic.com/en/api/messages:

  | Category | OpenAI (existing) | Anthropic (new) |
  | --- | --- | --- |
  | Reasoning | effort, summary | + thinking (enabled/disabled, budget_tokens) |
  | Tools | function, web_search, code_interpreter, tool_choice | ✓ web_search_20250305, bash_20250124, + text_editor (v1/v2/v3), tool_choice (auto/any/none/tool) |
  | Tool config | parallel_tool_calls | ✓ disable_parallel_tool_use |
  | Context | instructions, system message | ✓ system |
  | Caching | service_tier, prompt_cache_key | ✓ service_tier, + cache_control (ephemeral, ttl: 5m/1h) |
  | Metadata | metadata, safety_identifier | ✓ metadata, metadata.user_id |
  | Sampling | temperature, top_p, frequency_penalty, presence_penalty | ✓ temperature, top_p, + top_k |
  | Output control | stop, max_completion_tokens | ✓ stop_sequences, max_tokens |
  | Streaming | - | + stream |
  | Content types | - | + image (base64), document (text) |
  | Web search config | - | + allowed_domains, max_uses, user_location |

cross compatability coverage report -> https://github.com/braintrustdata/lingua/actions/runs/21009331032

PR #58, opened 2026-01-14
add chat completion parameter test cases

knjiang:
Add Chat Completions API parameter test cases to the existing `params.ts` file for bidirectional mapping coverage.

  New test coverage, got the request body from https://platform.openai.com/docs/api-reference/chat/create:

  | Category | Responses API (existing) | Chat Completions API (new) |
  | --- | --- | --- |
  | Reasoning | effort, summary | + reasoning_effort |
  | Text config | json_object, json_schema | ✓ response_format (both types) |
  | Tools | function, web_search, code_interpreter, specific choice, parallel_tool_calls | ✓ function tools, tool_choice, parallel_tool_calls |
  | Context | instructions, truncation | + system message (maps to instructions) |
  | Storage | store | ✓ store |
  | Caching | service_tier, prompt_cache_key | ✓ service_tier, prompt_cache_key |
  | Metadata | metadata, safety_identifier | ✓ metadata, safety_identifier |
  | Sampling | - | + temperature, top_p, frequency_penalty, presence_penalty, logprobs, top_logprobs |
  | Output control | - | + n (multiple completions), stop, max_completion_tokens |
  | Advanced | - | + prediction, seed, logit_bias |

cross compatability coverage report -> https://github.com/braintrustdata/lingua/actions/runs/21009331121

PR #54, opened 2026-01-14
add openai responses parameter test cases

knjiang:
Add parameter test cases for openai responses with the new `params.ts` file

new test coverage, got the request body from https://platform.openai.com/docs/api-reference/responses:

| Category | Before (existing) | After (new params.ts) |
| --- | --- | --- |
| Reasoning | effort (3 values), summary | ✓ (already covered) |
| Text config | verbosity | \+ json_object, json_schema |
| Tools | function type, auto choice | \+ web_search, code_interpreter, specific choice, parallel_tool_calls |
| Context | \- | \+ instructions, truncation |
| Storage | \- | \+ store |
| Caching | \- | \+ service_tier, prompt_cache_key |
| Metadata | \- | \+ metadata, safety_identifier |

cross compatability coverage report -> https://github.com/braintrustdata/lingua/actions/runs/21009330639

PR #61, opened 2026-01-16
Add universal param configs

knjiang:
## Summary
  This PR adds cross-provider compatability for chat completions, responses, anthropic.    

See -> https://github.com/braintrustdata/lingua/actions/runs/21328545339                                                                   
                                                                                                                               
### Parameter mappings                                                                                                   
                                                                                                                               
  | Feature | Chat Completions | Responses | Anthropic |                                                                       
  |---------|-----------------|-----------|-----------|                                                                        
  | Reasoning | `reasoning_effort` | `reasoning.effort` | `thinking.budget_tokens` |                                           
  | Structured output | `response_format.json_schema` | `text.format` | `output_format` |                                      
  | Tool selection | `tool_choice` | `tool_choice` | `tool_choice` + `disable_parallel_tool_use` |                             
  | Max tokens | `max_tokens` / `max_completion_tokens` | `max_output_tokens` | `max_tokens` |                                 
                                                                                                                               
### Testing

For each provider pair **(A → B)** across **Chat Completions / Responses / Anthropic**, we validate the deserialized Universal payload:

1. **Universal of source payload**
   - `U₁ = A payload → Universal`

2. **Translate across providers and re-canonicalize**
   - `U₂ = (A payload → Universal → B payload) → Universal`

3. **Diff the canonical forms**
   - Compare `U₁` vs `U₂`
   - Emit **field-level diffs** for any **lost / added / changed** fields

4. **Enforce in CI**
   - CI fails on any **unexpected** diffs

### Expected differences

- Known provider limitations / intentional lossy mappings are documented in `expected_differences.json`.
- Diffs covered by this file are treated as **allowed**; anything else is flagged as a **regression** and fails CI.


knjiang:
* **#70** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/70?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#74](https://github.com/braintrustdata/lingua/pull/74) <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/74?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#102](https://github.com/braintrustdata/lingua/pull/102) <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/102?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#72** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/72?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#69** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/69?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#61** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/61?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/61?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#59** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/59?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#58** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/58?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#54** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/54?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/lingua/src/providers/openai/adapter.rs
```diff
@@ -1,78 +1,40 @@
...
-    "stream",
-    // OpenAI-specific fields (not in UniversalParams) go to extras:
-    // stream_options, n, logprobs, top_logprobs, logit_bias,
-    // user, store, metadata, parallel_tool_calls, service_tier
-];
-
-/// Known request fields for OpenAI Responses API.
-/// These are fields extracted into UniversalRequest/UniversalParams.
-/// Fields not in this list go into `extras` for passthrough.
-const RESPONSES_KNOWN_KEYS: &[&str] = &[
```
knjiang, line 62: i move responses to `responses_adapter.rs`

crates/coverage-report/src/runner.rs
```diff
@@ -139,26 +256,65 @@ pub fn test_response_transformation(
...
-                    error: None,
-                },
+                Ok(target_universal) => {
+                    let target_universal_value = universal_response_to_value(&target_universal);
+                    let context = CompareContext::new(
+                        source_adapter.display_name(),
+                        target_adapter.display_name(),
+                        test_case,
+                    );
+                    let roundtrip_result = compare_values(
```
knjiang: i tightened the runner to be more accurate, basically we now compare:  
  
source -> universal 

with

source -> universal -> target -> universal

  
we basically deserialize universal -> JSON and do diffing

crates/coverage-report/src/expected_differences.json
```diff
@@ -0,0 +1,147 @@
+[
```
knjiang: this is how I'm storing expected differences

crates/coverage-report/src/runner.rs
```diff
@@ -56,8 +75,10 @@ pub fn test_request_transformation(
         None => {
             let error = format!("Source payload not found: {}", filename);
             return TransformResult {
-                level: ValidationLevel::MissingFixture,
```
remh, line 59: what's the reason we are removing the MissingFixture level? Seemed like it was handy to differentiate failure modes?

crates/lingua/src/universal/tool_choice.rs
```diff
@@ -0,0 +1,392 @@
...
+            }
+        }
+        Value::Object(obj) => {
+            let tool_name = obj.get("name").and_then(Value::as_str).map(String::from);
+
+            let mode = obj
+                .get("type")
+                .and_then(Value::as_str)
+                .and_then(|s| s.parse().ok())
+                .or(Some(ToolChoiceMode::Tool));
```
remh: if the object has an unrecognized type value (or no type at all), this defaults to Tool mode. This seems incorrect?

crates/lingua/src/universal/tools.rs
```diff
@@ -0,0 +1,902 @@
...
+// Batch Conversion Utilities
+// =============================================================================
+
+/// Convert a slice of UniversalTools to Anthropic format Value array.
+///
+/// Filters out tools that can't be converted (e.g., non-Anthropic builtins).
+pub fn tools_to_anthropic_value(tools: &[UniversalTool]) -> Option<Value> {
+    let converted: Vec<Value> = tools
+        .iter()
+        .filter_map(|t| t.to_anthropic_value().ok())
```
remh: looks like this would silently drop "bad" tool calls but i guess that's ok ?

crates/lingua/src/universal/request.rs
```diff
@@ -52,30 +96,529 @@ pub struct UniversalParams {
...
+// Stop Sequences Helper
+// =============================================================================
+
+/// Parse stop sequences from a JSON value.
+///
+/// Handles:
+/// - `"single_string"` → `vec!["single_string"]`
+/// - `["arr", "of", "strings"]` → `vec!["arr", "of", "strings"]`
+/// - Other types → `None`
+pub fn parse_stop_sequences(value: &Value) -> Option<Vec<String>> {
```
remh, line 581: I don't think this function has tests. we should add some

crates/lingua/src/universal/tools.rs
```diff
@@ -0,0 +1,902 @@
...
+                let mut obj = Map::new();
+                obj.insert("name".into(), Value::String(self.name.clone()));
+
+                if let Some(desc) = &self.description {
+                    obj.insert("description".into(), Value::String(desc.clone()));
+                }
+
+                obj.insert(
+                    "input_schema".into(),
+                    self.parameters.clone().unwrap_or_else(|| json!({})),
```
remh, line 291: should we only insert parameters if parameters is not empty instead?

crates/coverage-report/src/expected.rs
```diff
@@ -0,0 +1,285 @@
...
+        .map(|e| e.reason.clone())
+}
+
+/// Check if a transform error is expected for the given source→target translation.
+///
+/// Returns the reason if the error is expected (limitation), None if it's unexpected (a bug).
+/// Note: Errors containing "missing" are always treated as real failures, not limitations.
+pub fn is_expected_error(source: &str, target: &str, error_msg: &str) -> Option<String> {
+    // Errors about missing required fields are real failures, not limitations
+    if error_msg.contains("missing") {
```
remh: this seems overly broad? aren't there some limitations that would include the word missing?

remh (changes requested):

crates/coverage-report/src/runner.rs
knjiang: ya changing it to Skipped since its an expected error.

crates/coverage-report/src/expected.rs
knjiang: removed

crates/lingua/src/universal/tool_choice.rs
knjiang: ya will throw instead.

crates/lingua/src/universal/tools.rs
knjiang: i think we shoudl fail fast, gonna address.

crates/lingua/src/universal/tools.rs
knjiang: this is tool call schema, i think it has to be `{}` if empty

crates/coverage-report/src/normalizers.rs
```diff
@@ -0,0 +1,95 @@
...
+            normalize_user_content(content);
+        }
+        Message::Assistant { content, .. } => {
+            normalize_assistant_content(content);
+        }
+        Message::Tool { .. } => {}
+    }
+}
+
+fn normalize_user_content(content: &mut UserContent) {
```
knjiang, line 65: i couldn't find a good way to deserialize & compare UserContent since a single item text array and a string are semantically equivalent.  
  
this is the only normalization i do -> https://github.com/braintrustdata/lingua/blob/main/crates/lingua/src/universal/message.rs#L31  
  
i had an alternative approach of this code being normalized in the deserializer but that broke roundtrip tests so it lives here now.

crates/lingua/src/providers/openai/convert.rs
```diff
@@ -7,6 +7,38 @@ use crate::universal::{
...
+pub struct ChatCompletionResponseMessageExt {
+    #[serde(flatten)]
+    pub base: openai::ChatCompletionResponseMessage,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub reasoning: Option<String>,
+    /// Encrypted reasoning signature for cross-provider roundtrips (e.g., Anthropic's signature)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub reasoning_signature: Option<String>,
+}
+
```
knjiang, line 31: extended the chat completion type with reasoning + reasoning_signature.

i think reasoning_signature will be useful for ant/gemini. I think `reasoning` is supported via -> https://github.com/vllm-project/vllm/issues/27755 but `reasoning_signature` is smth i added alongside it.

crates/coverage-report/src/requests_expected_differences.json
```diff
@@ -0,0 +1,255 @@
...
+        { "pattern": "params.tool_choice", "reason": "Anthropic requires tool_choice to express disable_parallel_tool_use" }
+      ],
+      "errors": [
+        { "pattern": "does not support logprobs", "reason": "Anthropic doesn't support logprobs parameter" },
+        { "pattern": "does not support top_logprobs", "reason": "Anthropic doesn't support top_logprobs parameter" },
+        { "pattern": "does not support frequency_penalty", "reason": "Anthropic doesn't support frequency_penalty parameter" },
+        { "pattern": "does not support presence_penalty", "reason": "Anthropic doesn't support presence_penalty parameter" },
+        { "pattern": "does not support seed", "reason": "Anthropic doesn't support seed parameter" },
+        { "pattern": "does not support store", "reason": "Anthropic doesn't support store parameter" },
+        { "pattern": "does not support n > 1", "reason": "Anthropic doesn't support multiple completions" }
```
knjiang: this is something I was having trouble classifying whether we want to fail fast for unsupported parameters or silently drop when converting to a different model.  
  
For now, I've gone with fail-fast with completely unsupported params.

crates/coverage-report/tests/cross_provider_test.rs
```diff
@@ -0,0 +1,62 @@
...
+use coverage_report::types::TestFilter;
+use lingua::capabilities::ProviderFormat;
+use lingua::processing::adapters::adapters;
+
+/// Required providers for CI: Anthropic <-> ChatCompletions <-> Responses
+const REQUIRED_PROVIDERS: &[ProviderFormat] = &[
+    ProviderFormat::Responses,
+    ProviderFormat::OpenAI, // ChatCompletions
+    ProviderFormat::Anthropic,
+];
```
knjiang: The goal is to have this automatically run on every provider format, this is temporary while we incrementally make progress.

.github/workflows/cross-transformation-coverage.yml
```diff
@@ -45,18 +45,22 @@ jobs:
       - name: Generate coverage report
         run: |
           cargo run -p coverage-report > coverage_report.md
-        # This job is informational only - it always succeeds
-        # Click into the job summary to see the actual coverage report
 
       - name: Post coverage to job summary
+        if: always()
```
ankrgyl: what does `if: always()` do?

crates/lingua/src/universal/request.rs
```diff
@@ -52,30 +76,535 @@ pub struct UniversalParams {
...
 
-    /// Tool definitions (schema varies by provider)
-    pub tools: Option<Value>,
+    /// Whether to return log probabilities (OpenAI-specific but canonical)
+    pub logprobs: Option<bool>,
 
-    /// Tool selection strategy (varies by provider)
-    pub tool_choice: Option<Value>,
+    /// Number of top logprobs to return (0-20)
+    pub top_logprobs: Option<i64>,
```
ankrgyl, line 115: nit: if it's 1-20 can it be a smaller integer type like i8?

crates/lingua/src/universal/request.rs
```diff
@@ -52,30 +76,535 @@ pub struct UniversalParams {
...
+    /// Reasoning configuration for extended thinking / chain-of-thought.
+    ///
+    /// Uses canonical fields (`effort`, `budget_tokens`) for cross-provider conversion.
+    /// Skipped when disabled or empty to normalize `{enabled: false}` to `null`.
+    #[serde(skip_serializing_if = "reasoning_should_skip")]
+    pub reasoning: Option<ReasoningConfig>,
+
+    // === Metadata and identification ===
+    /// Request metadata (user tracking, experiment tags, etc.)
+    pub metadata: Option<Value>,
```
ankrgyl, line 150: what is this

crates/lingua/src/universal/request.rs
```diff
@@ -6,43 +6,67 @@ converted to/from any provider format.
...
-    pub extras: Map<String, Value>,
+    /// Provider-specific fields, keyed by the source ProviderFormat.
+    ///
+    /// When transforming back to the same provider, these extras are merged back.
+    /// When transforming to a different provider, they are ignored (no cross-pollination).
+    ///
+    /// Example: OpenAI Chat extras stay in `provider_extras[ProviderFormat::OpenAI]`
+    /// and are only merged back when converting to OpenAI Chat, not to Anthropic.
+    #[serde(skip)]
+    pub provider_extras: HashMap<ProviderFormat, Map<String, Value>>,
```
ankrgyl: it's slightly weird that this is not nested in `params` , at least to me. What as the rationale behind that?

crates/coverage-report/src/normalizers.rs
ankrgyl: just to clarify, this is just for testing right? if so it seems totally fine

crates/coverage-report/src/requests_expected_differences.json
ankrgyl: I think right now we silently drop. Maybe we should try this out on some other systems that do translation and see what they do?  
  
Uusally people solve stuff like this by having a flag and then propagating the choice to the user. Eg. mysql defaults to flexible type conversions and has a `strict` mode. Postgres is the opposite.

crates/coverage-report/tests/cross_provider_test.rs
ankrgyl: makes sense. Maybe leave this as a comment in the code

crates/lingua/src/providers/openai/convert.rs
ankrgyl: The reason it's added is so that other model providers than OpenAI who use chat completions can propagate it?

ankrgyl (commented):
Looks pretty good straightforward to me

- It's a little out of date, but it would be useful to write some typescript examples (eg in `examples/typescript/index.ts`) or even some rust examples that show the ergonomics of using parameters, so we can double check the format
- for parameters, i think it would be useful to write a "fuzz" style tester that for each provider, generates random values with respect to the openapi spec, and then roundtrips through `UniversalParams` (there is less entropy in parameters than raw requests, but this might just be generally useful)
- Is there a creative way we can port the test cases we have in the proxy/ repo? We have had a bunch of historical challenges with translating reasoning for example that is well captured in those tests.

crates/coverage-report/src/requests_expected_differences.json
knjiang: ran  


```bash
curl https://ai-gateway.vercel.sh/v1/chat/completions \                                                  
        -H "Authorization: Bearer <AI_GATEWAY_KEY>" \               
        -d '{                                                                                                   
          "model": "anthropic/claude-sonnet-4.5",                                                               
          "messages": [                                                                                         
            {                                                                                                   
              "role": "user",                                                                                   
              "content": "Write a one-sentence bedtime story about a unicorn."                                  
            }                                                                                                   
          ],                                                                                                    
          "stream": false,                                                                                      
          "logprobs": true,                                                                                     
          "top_logprobs": 3,                                                                                    
          "frequency_penalty": 0.5                                                                              
        }'
```

and the parameters were dropped

.github/workflows/cross-transformation-coverage.yml
knjiang: oops let me reorder. if always() just lets the CI continue if a step fails

crates/lingua/src/universal/request.rs
knjiang: responses/chatcompletions have it as:

```
metadata
map

Optional
Set of 16 key-value pairs that can be attached to an object. This can be useful for storing additional information about the object in a structured format, and querying for objects via API or the dashboard.

Keys are strings with a maximum length of 64 characters. Values are strings with a maximum length of 512 characters.
```

while anthrpoic has `metadata` as an object with one field `user_id`.  
  
https://platform.openai.com/docs/api-reference/responses/create#responses_create-metadata  
https://platform.claude.com/docs/en/api/messages

crates/coverage-report/src/requests_expected_differences.json
ankrgyl: Yeah then maybe we should silently drop, and offer a strict mode (this could even be a lingua/universal param) that enforces parameter translation is not dropped.

crates/lingua/src/universal/request.rs
ankrgyl: ahh ok. mind linking these in the comment

crates/coverage-report/src/requests_expected_differences.json
knjiang: https://www.notion.so/braintrustdata/Add-strict-mode-for-lingua-translation-2f4f7858028980c8a67ddc17ce88eb2d?source=copy_link <- deferring strict mode later

crates/lingua/src/universal/request.rs
knjiang: the openai generated type equivalent is i64 -> https://github.com/braintrustdata/lingua/blob/main/crates/lingua/src/providers/openai/generated.rs#L69

crates/lingua/src/universal/request.rs
knjiang: moved inside universal params -> https://github.com/braintrustdata/lingua/blob/add_universal_params_between_completions_responses_and_anthropic/crates/lingua/src/universal/request.rs#L174

ankrgyl:
> Looks pretty good straightforward to me
> 
> * It's a little out of date, but it would be useful to write some typescript examples (eg in `examples/typescript/index.ts`) or even some rust examples that show the ergonomics of using parameters, so we can double check the format
> * for parameters, i think it would be useful to write a "fuzz" style tester that for each provider, generates random values with respect to the openapi spec, and then roundtrips through `UniversalParams` (there is less entropy in parameters than raw requests, but this might just be generally useful)
> * Is there a creative way we can port the test cases we have in the proxy/ repo? We have had a bunch of historical challenges with translating reasoning for example that is well captured in those tests.

just to clarify, did you address these too?

crates/lingua/src/universal/reasoning.rs
```diff
@@ -0,0 +1,527 @@
...
+
+/// Convert token budget to effort level.
+///
+/// Uses ratio of budget/max_tokens with thresholds:
+/// - ratio < 0.35: low
+/// - 0.35 <= ratio < 0.65: medium
+/// - ratio >= 0.65: high
+pub fn budget_to_effort(budget: i64, max_tokens: Option<i64>) -> ReasoningEffort {
+    let max = max_tokens.unwrap_or(DEFAULT_MAX_TOKENS);
+    let ratio = budget as f64 / max as f64;
```
remh, line 145: Are we enforcing that `max` is a strictly positive integer?

crates/lingua/src/error.rs
```diff
@@ -9,12 +9,22 @@ pub enum ConvertError {
     #[error("Missing required field: {field}")]
     MissingRequiredField { field: String },
 
-    #[error("Invalid role: {role}")]
-    InvalidRole { role: String },
```
remh, line 13: FYI we still refer to this error in https://github.com/braintrustdata/lingua/blob/add_universal_params_between_completions_responses_and_anthropic/crates/lingua/docs/ADDING_PROVIDER_FORMAT.md#implementing-tryfromllm-conversions

crates/lingua/src/universal/tools.rs
```diff
@@ -0,0 +1,873 @@
...
+                    obj.insert("description".into(), Value::String(desc.clone()));
+                }
+
+                obj.insert(
+                    "parameters".into(),
+                    self.parameters.clone().unwrap_or_else(|| json!({})),
+                );
+
+                // Responses API function tools have strict: false by default
+                obj.insert("strict".into(), Value::Bool(false));
```
remh: what if a request had `strict: true` ? wouldn't that override it ?

crates/lingua/src/universal/response.rs
```diff
@@ -65,16 +68,303 @@ pub enum FinishReason {
...
+
+        match provider {
+            // OpenAI, Mistral, and Unknown use OpenAI format
+            ProviderFormat::OpenAI | ProviderFormat::Mistral | ProviderFormat::Unknown => {
+                let mut obj = serde_json::json!({
+                    "prompt_tokens": prompt,
+                    "completion_tokens": completion,
+                    "total_tokens": prompt + completion
+                });
+                let obj_map = obj.as_object_mut().unwrap();
```
remh: should we avoid using unwrap() and use expect() instead to give more context in case of crash? I think there are a few places in your PR where you use unwrap().

crates/lingua/src/universal/reasoning.rs
```diff
@@ -0,0 +1,527 @@
...
+pub const EFFORT_LOW_MULTIPLIER: f64 = 0.25;
+
+/// Multiplier for "medium" effort (50% of max_tokens)
+pub const EFFORT_MEDIUM_MULTIPLIER: f64 = 0.50;
+
+/// Multiplier for "high" effort (75% of max_tokens)
+pub const EFFORT_HIGH_MULTIPLIER: f64 = 0.75;
+
+/// Threshold below which budget is considered "low" effort
+pub const EFFORT_LOW_THRESHOLD: f64 = 0.35;
```
remh, line 71: are those thresholds documented anywhere? or is it just something you came up with (which would be fine)

remh (changes requested):
A few comments that need to be addressed but otherwise it looks good to me.

crates/lingua/src/universal/reasoning.rs
knjiang: these are copied from the old proxy as a starting point.

crates/lingua/src/universal/tools.rs
knjiang: chat completions don't have strict;true for tools. google and anthropic does though so i'll preemptively add support now.  
  
i realized our anthropic spec is out of date

crates/lingua/src/universal/reasoning.rs
knjiang: good catch! adddedtests

knjiang:
> * It's a little out of date, but it would be useful to write some typescript examples (eg in `examples/typescript/index.ts`) or even some rust examples that show the ergonomics of using parameters, so we can double check the format
> * for parameters, i think it would be useful to write a "fuzz" style tester that for each provider, generates random values with respect to the openapi spec, and then roundtrips through `UniversalParams` (there is less entropy in parameters than raw requests, but this might just be generally useful)
> * Is there a creative way we can port the test cases we have in the proxy/ repo? We have had a bunch of historical challenges with translating reasoning for example that is well captured in those tests.

1. deferring to https://github.com/braintrustdata/lingua/pull/69
2. kk, i'll separate out the fuzz tester?
3. done in this PR by pulling in the test cases -> https://github.com/braintrustdata/lingua/pull/61/changes/f1340cbcc805b1ede43035063eb10f2af8826d0a

there are still some proxy failures but i got most tests passing.
```
proxyAnthropicAudioError - audio PR
proxyOpenAIO3MiniReasoning - openai capabilitiescapabilities
proxyGoogleReasoning - google next PR
proxyGoogleAudioSupport -  google next PR
proxyGoogleVideoSupport - google next PR
proxyAzureParamFiltering - capabilities
proxyOpenAIO3MiniStreamingReasoning - openai capabilities
proxyOpenAIPdfUrlConversion - I'm not sure how to do this, we would have to make a network call and then encode(?)
```

payloads/proxy/cases.ts
```diff
@@ -0,0 +1,915 @@
+import OpenAI from "openai";
```
knjiang, line 1: all the cases i ported from proxy that "fit" into this. some of the proxy cases intercepted calls so that doesn't really work yet. I have a subsequent PR with a more "intercept"-like approach. https://app.graphite.com/github/pr/braintrustdata/lingua/72  
  
there are still some proxy failures but i got most tests passing.

```
proxyAnthropicAudioError - audio PR
proxyOpenAIO3MiniReasoning - openai capabilitiescapabilities
proxyGoogleReasoning - google next PR
proxyGoogleAudioSupport -  google next PR
proxyGoogleVideoSupport - google next PR
proxyAzureParamFiltering - capabilities
proxyOpenAIO3MiniStreamingReasoning - openai capabilities
proxyOpenAIPdfUrlConversion - I'm not sure how to do this, we would have to make a network call and then encode(?)

```

remh (approved):

PR #65, opened 2026-01-23
Fix parsing for anthropic + ai sdk case

ankrgyl:
The test captures the case


knjiang (approved):

knjiang:
i think linting just needs to be fixed

PR #71, opened 2026-01-28
update bedrock runtime

ibolmo:


ibolmo:
going to close this.. may actually not have been necessary

PR #78, opened 2026-02-04
Force translation for openai reasoning models

knjiang:
chat completions can have `max_tokens` as a parameter for the API but it isn't accepted for certain models, see error:
```
  "error": {
    "message": "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
    "type": "invalid_request_error",
    "param": "max_tokens",
    "code": "unsupported_parameter"
  }
}

Caused by:
    HTTP 400 Bad Request: {
      "error": {
        "message": "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
        "type": "invalid_request_error",
        "param": "max_tokens",
        "code": "unsupported_parameter"
      }
```
all models associated with chat completions api accept `max_completion_tokens` which is why when we transform back from universal, we always insert `max_completion_tokens` ->  https://github.com/braintrustdata/lingua/blob/main/crates/lingua/src/providers/openai/adapter.rs#L209

This PR adds forcing translation automatically for these reasoning models that don't support `max_tokens`

### DIFF
--- a/crates/lingua/src/processing/transform.rs
+++ b/crates/lingua/src/processing/transform.rs
@@ -511,9 +511,9 @@ fn needs_forced_translation(payload: &Value, model: Option<&str>, target: Provid
     {
         // If the model doesn't support max_tokens, we need to force translation
         let request_model = payload.get("model").and_then(Value::as_str).or(model);
-        return request_model
+        request_model
             .map(|m| !model_supports_max_tokens(m))
-            .unwrap_or(false);
+            .unwrap_or(false)
     }
 
     #[cfg(not(feature = "openai"))]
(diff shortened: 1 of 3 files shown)


crates/lingua/src/processing/transform.rs
```diff
@@ -497,6 +501,23 @@ fn detect_adapter(
...
+/// Reasoning models (gpt-5.*, o1-*, etc.) require `max_completion_tokens` instead of `max_tokens`.
+fn needs_forced_translation(payload: &Value, model: Option<&str>, target: ProviderFormat) -> bool {
+    if target != ProviderFormat::OpenAI {
+        return false;
+    }
+
+    #[cfg(feature = "openai")]
+    {
+        let request_model = payload.get("model").and_then(Value::as_str).or(model);
+        return request_model.map(is_reasoning_model_name).unwrap_or(false);
```
clutchski: do you think this a stable way of connecting max_tokens ? i think i might prefer `model_supports_max_completion_tokens` (or wahtever)

knjiang: sure ill decouple the lists

clutchski (approved):

PR #80, opened 2026-02-05
Refactor openai capabilities

knjiang:
Deleted a bunch of dead code and added support for stripping temperature.

### DIFF
--- a/crates/lingua/src/processing/transform.rs
+++ b/crates/lingua/src/processing/transform.rs
@@ -17,7 +17,7 @@ use crate::capabilities::ProviderFormat;
 use crate::error::ConvertError;
 use crate::processing::adapters::{adapter_for_format, adapters, ProviderAdapter};
 #[cfg(feature = "openai")]
-use crate::providers::openai::model_supports_max_tokens;
+use crate::providers::openai::model_needs_transforms;
 use crate::serde_json::Value;
 use crate::universal::{UniversalResponse, UniversalStreamChunk};
 use thiserror::Error;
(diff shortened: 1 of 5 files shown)


crates/lingua/src/processing/transform.rs
```diff
@@ -509,11 +509,9 @@ fn needs_forced_translation(payload: &Value, model: Option<&str>, target: Provid
 
     #[cfg(feature = "openai")]
     {
-        // If the model doesn't support max_tokens, we need to force translation
+        // Force translation if model needs any transforms (temperature stripping, max_tokens conversion, etc.)
         let request_model = payload.get("model").and_then(Value::as_str).or(model);
-        request_model
-            .map(|m| !model_supports_max_tokens(m))
```
knjiang, line 515: basically we force translation in order for "capabilities" to apply

crates/lingua/src/providers/openai/responses_adapter.rs
```diff
@@ -233,11 +226,7 @@ impl ProviderAdapter for ResponsesAdapter {
                 .map_err(|e| TransformError::SerializationFailed(e.to_string()))?,
         );
 
-        // Pass temperature through for non-reasoning models
-        // Reasoning models (o1-*, o3-*, gpt-5-*) don't support temperature
-        if !is_reasoning_model(model) {
-            insert_opt_f64(&mut obj, "temperature", req.params.temperature);
-        }
```
knjiang, line 240: we already supported this for responses api - im using the same abstraction instead `apply_model_transforms` for both responses/chat completions since they use same model family

crates/lingua/src/processing/transform.rs
knjiang: for context i added -> https://github.com/braintrustdata/lingua/commit/deb50d631544c96619f673b0bf72a463b7609d46

crates/lingua/src/providers/openai/capabilities.rs
```diff
@@ -1,140 +1,62 @@
...
+pub fn apply_model_transforms(model: &str, obj: &mut Map<String, Value>) {
+    for transform in get_model_transforms(model) {
+        match transform {
+            StripTemperature => {
+                obj.remove("temperature");
+            }
+            ForceMaxCompletionTokens => {
+                // (Responses API) max_output_tokens is valid.
+                if obj.contains_key("max_output_tokens") {
+                    return;
```
remh: shouldn't this be `continue` instead of `return` ?

crates/lingua/src/providers/openai/capabilities.rs
remh, line 1: Let's add some unit tests to the functions here

remh (changes requested):

remh (approved):

PR #82, opened 2026-02-05
Push Bedrock Anthropic payload mutations into lingua

clutchski:
## Summary

- Move Bedrock Anthropic payload mutations (strip `model`/`stream`, inject `anthropic_version`) into lingua's `transform_request` -- the router just calls `transform_request`, gets `Bytes` back, done
- Change Bedrock Anthropic models from `"format": "converse"` to `"format": "anthropic"` in catalog
- Use `spec.format` instead of `provider.format()` in the router so Bedrock Anthropic models get transformed to Anthropic format
- Add `add_provider_for_format()` and `add_model_override()` on `RouterBuilder` so Bedrock can serve both Converse and Anthropic format models
- Add `bedrock_messages_event_stream` for Bedrock Anthropic streaming (base64-encoded AWS binary event stream)
- Add `invoke_url` format-aware routing (`/invoke` for Anthropic, `/converse` for Converse)
- No `mutate_payload`, no `transform_request_to_value`, no `TransformResultValue`

## Test plan

- [x] Unit tests for `is_bedrock_anthropic_model` detection
- [x] Unit test verifying mutations applied (model/stream stripped, anthropic_version injected, max_tokens preserved)
- [x] Unit test verifying non-Bedrock Anthropic models still passthrough
- [x] Bedrock invoke_url tests (Converse vs Anthropic endpoints)
- [x] Bedrock Messages event stream base64 decoding tests
- [x] Router multi-format provider test (OpenAI/Anthropic/Converse through same provider)
- [x] `cargo test` passes across full workspace
- [x] `cargo check` passes for gateway

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/lingua/src/processing/transform.rs
+++ b/crates/lingua/src/processing/transform.rs
@@ -508,7 +508,7 @@ fn detect_adapter(
 
 /// Check if a model name indicates a Bedrock Anthropic model
 /// (e.g. `us.anthropic.claude-3-5-sonnet-20241022-v2:0` or `anthropic.claude-3-haiku-20240307-v1:0`).
-fn is_bedrock_anthropic_model(model: Option<&str>) -> bool {
+pub fn is_bedrock_anthropic_model(model: Option<&str>) -> bool {
     model.is_some_and(|m| m.starts_with("anthropic.") || m.contains(".anthropic."))
 }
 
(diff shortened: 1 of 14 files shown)


crates/lingua/src/processing/transform.rs
```diff
@@ -231,30 +235,101 @@ pub fn transform_request(
...
+        bytes: Bytes::from(bytes),
+        source_format,
+    })
+}
+
+/// Result of a value-based transformation operation.
+///
+/// Like `TransformResult` but returns a `Value` instead of `Bytes`,
+/// allowing callers to mutate the payload before serialization.
+#[derive(Debug, Clone)]
```
knjiang: we shouldn't apply the transform here, we should just put it in the anthropic folder

PR #100, opened 2026-02-13
Add event_type to StreamChunk for SSE event line preservation

clutchski:
The Anthropic SDK requires SSE `event:` lines to dispatch streaming events. Embed the event type from upstream SSE parsing into StreamChunk so downstream consumers can emit it without re-parsing JSON.

Also moves bedrock parsing lower so it doesn't interpret anthropic payloads as bedrock andthropic


knjiang (approved):

knjiang:
i think the linter is failing

PR #101, opened 2026-02-13
LLM router: add API key support for Bedrock provider

aswink:


crates/braintrust-llm-router/src/providers/bedrock.rs
```diff
@@ -115,9 +115,18 @@ impl BedrockProvider {
     }
 
     fn sign_request(&self, url: &Url, body: &[u8], auth: &AuthConfig) -> Result<HeaderMap> {
-        let (access_key, secret_key, session_token, region, service) = auth
-            .aws_credentials()
-            .ok_or_else(|| Error::Auth("AWS credentials required for Bedrock".into()))?;
+        if let AuthConfig::ApiKey { .. } = auth {
+            // TODO: should this instead be: `let mut headers =
+            // self.build_headers(client_headers)`?
```
knjiang: ya i think self.build_headers() is more inline with the other providers so we can forward client headers (maybe they have some special header they want to use)

knjiang (approved):

PR #108, opened 2026-02-14
Fuzz test anthropic + responses

ankrgyl:
Add

anthropic->universal->anthropic
chat->universal->anthropic->universal->chat


crates/lingua/src/providers/anthropic/adapter.rs
```diff
@@ -36,6 +38,41 @@ use crate::universal::{
 /// Default max_tokens for Anthropic requests (matches legacy proxy behavior).
 pub const DEFAULT_MAX_TOKENS: i64 = 4096;
 
+fn system_to_user_content(system: System) -> UserContent {
```
knjiang: hm... why cna't this be in the anthropic [convert.rs](http://convert.rs)? isn't this part of the message payload?

ankrgyl: good point. moved

knjiang (approved):

PR #123, opened 2026-02-23
Handle anthropic system messages

ankrgyl:
Technically, Anthropic message arrays don't contain system role'd messages (it's a separate parameter), so parsing them as an Anthropic message array does not work.

This change adds a special anthropic+system message type so we can parse this case more effectively.


payloads/import-cases/span_cfce074d2e284f8f.assertions.json
```diff
@@ -0,0 +1,13 @@
+{
```
CLowbrow: give this a more descriptive filename before merging?

CLowbrow (approved):
seems fine but I suck at rust

PR #125, opened 2026-02-24
[Anthropic] Fix case where anthropic models have tool choice alongside reasoning

knjiang:
Before this diff, `chat_completions_to_anthropic` would emit:

```json
{ 
    "error": "400 {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"Thinking may not be enabled when tool_choice forces tool use.\"},\"request_id\":\"req_011CYTRzwy2T1KfQDz2t7Xg4\"}", "name": "Error" }
```

### DIFF
--- a/payloads/scripts/anonymize-json.test.ts
+++ b/payloads/scripts/anonymize-json.test.ts
@@ -1,19 +1,9 @@
 import { describe, expect, it } from "vitest";
 import { anonymizeJsonValue, type JsonValue } from "./anonymize-json";
 
-interface DefaultFixture {
-  input: Array<{
-    role: string;
-    content: string | Array<{ type: string; text: string }>;
-    id?: string;
-    finish_reason?: string;
-  }>;
-  metadata: string;
-}
-
 describe("anonymizeJsonValue", () => {
   it("anonymizes content and metadata strings by default", () => {
-    const input: DefaultFixture = {
+    const input: JsonValue = {
       input: [
         { role: "user", content: "hello world", id: "user-1" },
         {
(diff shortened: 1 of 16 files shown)


payloads/scripts/anonymize-json.ts
knjiang, line 1: drive-by: fix-lint

payloads/scripts/anonymize-json.test.ts
knjiang, line 1: drive-by: fix-lint

payloads/snapshots/toolChoiceRequiredWithReasoningParam/chat-completions/followup-request.json
```diff
@@ -0,0 +1,59 @@
+{
```
knjiang, line 1: snapshots are generated straight from `params.ts`

payloads/transforms/chat-completions_to_anthropic/toolChoiceRequiredWithReasoningParam.json
```diff
@@ -0,0 +1,33 @@
+{
```
knjiang, line 1: transforms are generated bby taking the payload of `params.ts` of an openai test case and just swapping the model in for an anthropic/gemini model  
  
they are then recorded in directory of `chat_completions_to_x`

aswink (approved):
LGTM, though I wonder if choosing between tool choice and reasoning should be an option somewhere. It's not clear to me that preferring tool choice is always what users will want

PR #127, opened 2026-02-26
Remove retries in LLM router

aswink:
lingua is more intended to do provider translation than request/retry
semantics, so remove retries from this library. Callers can take care of
retries themselves.


aswink:
* **#136** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#127** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/braintrust-llm-router/src/client.rs
```diff
@@ -18,10 +18,13 @@ pub struct ClientSettings {
 impl Default for ClientSettings {
     fn default() -> Self {
         Self {
-            connect_timeout: Duration::from_secs(10),
-            request_timeout: Duration::from_secs(300),
+            // Similar to what anthropic and openai SDKs use
```
aswink, line 21: Not really related to the rest of the change, happy to drop it if it seems unnecessary. Though I do wonder if we'll need the higher request timeout

crates/braintrust-llm-router/src/router.rs
```diff
@@ -147,25 +140,16 @@ impl Router {
...
         let payload = match lingua::transform_request(body.clone(), format, Some(&spec.model)) {
             Ok(TransformResult::PassThrough(bytes)) => bytes,
             Ok(TransformResult::Transformed { bytes, .. }) => bytes,
             Err(TransformError::UnsupportedTargetFormat(_)) => body.clone(),
             Err(e) => return Err(e.into()),
         };
 
-        let response_bytes = self
-            .execute_with_retry(
-                provider.clone(),
```
knjiang, line 164: im a bit wary of incurring the cost of `lingua::transform_request` on every retry. maybe we can transform up-front? or pass through a retry policy?

knjiang (commented):
just one comment to think through. i'm not sure if i totally like passing a retry policy throiugh but that was my first thought

crates/braintrust-llm-router/src/router.rs
aswink: Good point, maybe we can transform up front? lmk what you think of the latest version

crates/braintrust-llm-router/src/router.rs
```diff
@@ -100,16 +96,40 @@ type ResolvedRoute<'a> = (
...
     Arc<ModelSpec>,
     ProviderFormat,
-    RetryStrategy,
 );
 
+pub struct CompletionRequest {
+    /// The payload to send to the provider.
+    payload: Bytes,
+    /// The format of the request payload.
+    request_format: ProviderFormat,
```
knjiang: [nit] maybe it makes sense to call this `model_format` or `provider_format`. this is the format of whatever provider was resolved from the model name. 

i'm leaning `provider_format` since a provider can have multiple formats and the `provider.complete()` API currently uses the value.

knjiang (approved):

crates/braintrust-llm-router/src/router.rs
aswink: provider_format makes sense to me too, though it is a little confusing that the type is called ProviderFormat too

knjiang: yea maybe request_format is fine. just want to make sure we don't imply that it's the format of the original request payload.

aswink:
CI passes on the main repo: https://github.com/braintrustdata/braintrust/pull/11253

PR #129, opened 2026-02-27
Improve import logic for different output shapes

ankrgyl:
You can output a list of messages, a single message (which itself may be a list of content parts), or just text. This change fixes the logic so that we try all of these in a provider-agnostic way (though the specific bug has to do with Anthropic)


CLowbrow (approved):
clippy is mad at you but seems fine otherwise

PR #136, opened 2026-03-04
Add support for available_providers in model spec

aswink:
Without this, we can only guess which providers can run a model. For
example, it's obvious that a vertex format model would run on GCP
Vertex, but what about a model like llama3? It could run on any provider.
This change adds explicit mappings from model ID -> supported providers.

This allows callers to specify which providers a model is available on. If
empty, then we fallback to the existing behavior. Also update
provider_aliases so it follows the same logic as resolve_providers,
particularly that it returns providers in the same order.

Depends on https://github.com/braintrustdata/braintrust-proxy/pull/407

### DIFF
--- a/crates/braintrust-llm-router/src/catalog/resolver.rs
+++ b/crates/braintrust-llm-router/src/catalog/resolver.rs
@@ -28,7 +28,7 @@ impl ModelResolver {
         Arc::clone(&self.catalog)
     }
 
-    pub fn resolve(&self, model: &str) -> Result<(Arc<ModelSpec>, ProviderFormat, String)> {
+    pub fn resolve(&self, model: &str) -> Result<(Arc<ModelSpec>, ProviderFormat, Vec<String>)> {
         let spec = self
             .catalog
             .get(model)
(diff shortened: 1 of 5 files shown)


aswink:
* **#136** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/136?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#142** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

crates/braintrust-llm-router/src/catalog/resolver.rs
```diff
@@ -44,14 +44,20 @@ impl ModelResolver {
...
-        Ok((spec, format, provider_alias))
+        let provider_aliases = self
+            .aliases
+            .get(model)
+            .map(|alias| vec![alias.clone()])
+            .unwrap_or_else(|| {
+                if !spec.available_providers.is_empty() {
+                    spec.available_providers.clone()
+                } else if lingua::is_vertex_model(model) {
+                    vec!["vertex".to_string()]
```
knjiang, line 55: i had to do this because the current model_list.json doesn't tell u that its provider -> https://github.com/braintrustdata/braintrust-proxy/blob/main/packages/proxy/schema/model_list.json#L4022

i wonder if we can simplify now that we have available providers?

this could be a follow-up i tackle to get more familiar with the code.

knjiang (approved):
looks good no blockers

crates/braintrust-llm-router/src/catalog/resolver.rs
aswink: Oh, in that case yes? I'll leave that to a follow up just in case, though.

aswink:
### Merge activity

* **Mar 14, 4:44 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/braintrustdata/lingua/136).
* **Mar 14, 4:45 AM UTC**: [Graphite](https://app.graphite.com/github/pr/braintrustdata/lingua/136) rebased this pull request as part of a merge.
* **Mar 14, 4:54 AM UTC**: @aswink merged this pull request with [Graphite](https://app.graphite.com/github/pr/braintrustdata/lingua/136).

PR #146, opened 2026-03-19
Fix using anthropic models with responses API in AI SDK

aswink:
A customer ran into an issue where AI SDK was failing because it has strict
type checks using a zod format. It expects responses API to always have IDs for
items and an annotations field, even if empty. We were dropping these when
translating between providers. Fix this by restoring the ID if the format
matches, otherwise putting a placeholder in. Also always fill in an empty
annotations.

At some point in the future we should look at adding annotations (openai) and
citations (anthropic's version) to the universal format.

### DIFF
--- a/crates/braintrust-llm-router/src/providers/anthropic.rs
+++ b/crates/braintrust-llm-router/src/providers/anthropic.rs
@@ -124,35 +124,23 @@ impl crate::providers::Provider for AnthropicProvider {
     ) -> Result<Bytes> {
         let url = self.messages_url();
 
-        #[cfg(feature = "tracing")]
-        tracing::debug!(
-            target: "bt.router.provider.http",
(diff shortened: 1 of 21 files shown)


aswink:
* **#146** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/146?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

knjiang (approved):

knjiang:
i added a hook to validate via the AI SDK zod schema tests failed before :

image

tests pass after:

image

PR #147, opened 2026-03-19
Fix vertex endpoint for newer models and responses schema 

knjiang:


crates/lingua/src/providers/openai/responses_adapter.rs
```diff
@@ -486,6 +486,7 @@ impl ProviderAdapter for ResponsesAdapter {
         map.insert("tool_choice".into(), Value::String("none".into()));
         map.insert("tools".into(), Value::Array(vec![]));
         map.insert("parallel_tool_calls".into(), Value::Bool(false));
+        map.insert("incomplete_details".into(), Value::Null);
```
knjiang, line 489: drive-by fix to allow AI SDK

crates/braintrust-llm-router/src/providers/vertex.rs
```diff
@@ -98,9 +111,50 @@ impl VertexProvider {
...
+            if !extra.locations.is_empty() {
+                let idx = rand::thread_rng().gen_range(0..extra.locations.len());
+                return extra.locations[idx].clone();
+            }
+        }
+        if !self.config.location.is_empty() {
+            return self.config.location.clone();
+        }
+        default_location.to_string()
+    }
```
knjiang, line 128: this matches the proxy code where we randomly choose location from the locations array for people who specify custom models with a variety of lcoations -> https://github.com/braintrustdata/braintrust-proxy/blob/main/packages/proxy/src/proxy.ts#L2451

clutchski (approved):

PR #149, opened 2026-03-20
Optimize build

ibolmo:
Part of: https://github.com/braintrustdata/braintrust/pull/11761

Trying to speed up our builds


bindings/lingua-wasm/turbo.json
```diff
@@ -3,12 +3,15 @@
...
-  "globalDependencies": [
-    "../../crates/lingua/src/**/*.rs",
-    "../../crates/lingua/Cargo.toml"
-  ],
   "tasks": {
     "build": {
+      "inputs": [
+        "src/**",
+        "package.json",
+        "../../crates/lingua/src/**/*.rs",
```
ibolmo, line 11: according to claude using input vs. globalDependencies are better because globalDependencies is a "sledgehammer" 🤷🏻 

this adds in package.jsona adn src/** which wasn't there before

ibolmo:
going to close this one out for now

PR #154, opened 2026-03-25
Remove default retries from llm router

knjiang:
This PR removes default retries from llm router. we don't document anywhere in https://www.braintrust.dev/docs/deploy/gateway#use-the-braintrust-gateway about retry so i think its fine until we figure out a better way to do retries for our svc. 

i think retries + provider fallbacks should be designed together in the future


knjiang:
* **#154** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/154?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/154?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/braintrust-llm-router/src/router.rs
```diff
@@ -285,23 +285,27 @@ impl Router {
...
+                    model = %model,
+                    output_format = ?output_format,
+                    all_aliases = ?aliases,
+                    spec = ?spec,
+                    catalog_format = ?catalog_format,
+                    error = %error,
+                    "failed to resolve any provider aliases",
+                );
+            }
+        }
```
knjiang, line 308: drive-by: we have a ton of warning logs that aren't necessary.

we should only error if we failed to resolve any provider aliases.

cpinn (approved):

PR #160, opened 2026-03-30
[Bedrock] Translate image urls to base64 images

cpinn:
### Background 

A customer is running into an issue calling bedrock anthropic with image url.

Image url is not supported by bedrock anthropic even though anthropic now supports image url. (ie. using  "model": "claude-haiku-4-5-20251001" works fine but bedrock anthropic will error).

This request works with the old proxy:
```
curl -X POST https://api.braintrust.dev/v1/proxy/chat/completions \
  -H "Authorization: Bearer <TOKEN>" \ 
  -d '{
    "model": "us.anthropic.claude-haiku-4-5-20251001-v1:0",
    "max_tokens": 100,
    "messages": [
      {
        "role": "user",
        "content": [
          {"type": "text", "text": "What is in this image?"},
          {"type": "image_url", "image_url": {"url": <any image url>}}
        ]
      },
      {"role": "assistant", "content": "It looks like a kitten."},
      {"role": "user", "content": "Describe it again briefly."}
    ]
```
fails with the new proxy at gateway.braintrust.dev with the error below
```
{"message":"messages.0.content.1.image.source.base64.data: URL sources are not supported"}% 
```
Preserve old proxy behavior of re-writing image_url into base64 format which is accepted by [Bedrock Anthropic](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ImageSource.html)

Converse also does not support image_url.

### Fixes

Prior to translating the requests to external providers download the urls in order to send the request in base64 format.

### DIFF
--- a/crates/braintrust-llm-router/src/router.rs
+++ b/crates/braintrust-llm-router/src/router.rs
@@ -1,6 +1,4 @@
 use std::collections::HashMap;
-use std::future::Future;
-use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::time::sleep;
@@ -16,10 +14,7 @@ use crate::error::{Error, Result};
 use crate::providers::{ClientHeaders, Provider};
 use crate::retry::{RetryPolicy, RetryStrategy};
 use crate::streaming::{transform_stream, ResponseStream};
-use lingua::processing::{adapter_for_format, adapters};
 use lingua::serde_json::Value;
-use lingua::universal::message::{Message, UserContent, UserContentPart};
-use lingua::util::media::MediaBlock;
 use lingua::ProviderFormat;
 use lingua::{TransformError, TransformResult};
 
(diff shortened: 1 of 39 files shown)


crates/braintrust-llm-router/src/router.rs
```diff
@@ -157,12 +157,9 @@ impl Router {
             .first()
             .ok_or_else(|| Error::NoProvider(output_format))?;
         let (_, provider, auth, spec, format, strategy) = route;
-        let payload = match lingua::transform_request(body.clone(), *format, Some(&spec.model)) {
-            Ok(TransformResult::PassThrough(bytes)) => bytes,
-            Ok(TransformResult::Transformed { bytes, .. }) => bytes,
-            Err(TransformError::UnsupportedTargetFormat(_)) => body.clone(),
```
knjiang, line 163: feels like a lot of indirection with defautl_prepare_reqeust and prepare_request, can we go back to inlining?

knjiang: ah i see we need to do custom stufff for bedrock.

cpinn: The other option which was the first change I had been testing with (and also works)

is running the translation for anything just above this payload request transform and inline the request urls.

It was a lot less code but slightly less provider specific and I am uncertain if we will have any of these other provider quirks. 

open to either fix though.

knjiang: i thnk we can go for the less code approach first and if we start seeing a pattern transition to prepare payload style

crates/lingua/src/providers/anthropic/convert.rs
```diff
@@ -166,24 +166,50 @@ impl TryFromLLM<generated::InputMessage> for Message {
...
+                                                    generated::FluffyMediaType::ImageGif => "image/gif".to_string(),
+                                                    generated::FluffyMediaType::ImageWebp => "image/webp".to_string(),
+                                                    generated::FluffyMediaType::ApplicationPdf => "application/pdf".to_string(),
+                                                    generated::FluffyMediaType::TextPlain => "text/plain".to_string(),
+                                                });
+                                                let source_type = purple_source.source_type.clone();
+                                                let image = match source_type {
+                                                    generated::FluffyType::Base64 => {
+                                                        purple_source.data.ok_or_else(|| {
+                                                            ConvertError::MissingRequiredField {
```
cpinn: cc: @knjiang I reverted the changes in this anthropic file and in the bedrock file but I am curious, do you think we should add more errors in the provider if inputs aren't correct? ie. if image_url/url type you are a url and if base64 you have data etc.

crates/braintrust-llm-router/src/router.rs
```diff
@@ -107,6 +112,144 @@ type ResolvedRoute<'a> = (
     RetryStrategy,
 );
 
+fn is_remote_image_url(value: &str) -> bool {
+    value.starts_with("http://") || value.starts_with("https://")
+}
+
+const BEDROCK_REMOTE_MEDIA_MAX_BYTES: usize = 5 * 1024 * 1024;
```
cpinn: should we carry this limit over: https://github.com/braintrustdata/braintrust-proxy/blob/376d1eaf5b0cb81f73566e914db062d2019363bb/packages/proxy/src/providers/bedrock.ts#L432?

crates/braintrust-llm-router/src/router.rs
```diff
@@ -107,6 +112,144 @@ type ResolvedRoute<'a> = (
...
+    matches!(
+        format,
+        ProviderFormat::BedrockAnthropic | ProviderFormat::Converse
+    )
+}
+
+async fn inline_remote_image_urls_with_fetch<F>(
+    request: &mut lingua::UniversalRequest,
+    mut fetch: F,
+) -> Result<()>
```
knjiang: lets move these functions in bedrock.rs since it's only belonging there right? i think we should also add a comment explaining how we pre-process bedrock payloads to inline fetch images.

knjiang (approved):
lgtm, no real big comments. im assuming we ran payloads/proxy/cases.ts?

knjiang (commented):
wait actually i think we want the logic in convert.rs. i was just saying we should hard-code the pre-processing.

crates/lingua/src/providers/anthropic/convert.rs
knjiang: wait why are we able to get rid of this convert.rs change?

cpinn: discussed offline. 

It is unclear to me if we should be (as I did in this changeset), wrap/modify potential errors in the converter and throw our own error OR if we should continue to re-surface 400s from the underlying provider. 

ie. if you send the http url to bedrock it responds with a 400 already. Would it be useful to check prior to sending and still send something like "Bedrock requires images to be sent as a base64 encoded object"? It was clear that would really help as it could obscure the underlying error.

cpinn:
> wait actually i think we want the logic in convert.rs. i was just saying we should hard-code the pre-processing.

@knjiang lets chat tomorrow about the synchronous lingua interface and if it does make more sense to have the async prepare_request from the first commit or modify lingua with an asynchronous method to handle something like this.

cpinn:
We have decided to merge this as our goal of "providing a universal translator" means that something needs to be done when a user passes a image_url and that is not a supported format. 

We have moved over the behavior of changing the request from an image_url into a base64 image but we likely will want to re-visit the interface here as some of the lingua conversion is now tied to the bedrock provider.

PR #161, opened 2026-03-30
google vertex models resolve to google format regardless of model spec format

knjiang:
To align behavior with old proxy, we need to always resolve google vertex models to the google format.  
  
previously, the code would just think to passthrough chat completions and error out since the output format and api format was chat completions

### DIFF
--- a/crates/braintrust-llm-router/src/catalog/resolver.rs
+++ b/crates/braintrust-llm-router/src/catalog/resolver.rs
@@ -41,6 +41,11 @@ impl ModelResolver {
             && lingua::is_vertex_anthropic_model(model)
         {
             ProviderFormat::VertexAnthropic
+        } else if lingua::is_vertex_google_model(model) {
+            // Force Google format regardless of what the spec says.
+            // Custom models may declare format:openai but Vertex's Generative AI
+            // API requires the Google generateContent wire format.
+            ProviderFormat::Google
         } else {
             spec.format
         };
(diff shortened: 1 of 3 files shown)


knjiang:
* **#161** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/braintrust-llm-router/src/catalog/resolver.rs
```diff
@@ -41,6 +41,11 @@ impl ModelResolver {
             && lingua::is_vertex_anthropic_model(model)
         {
             ProviderFormat::VertexAnthropic
+        } else if lingua::is_vertex_google_model(model) {
+            // Force Google format regardless of what the spec says.
```
cpinn, line 45: So if the user had declared `format: google` this wouldn't have been an issue right?

cpinn (approved):

crates/braintrust-llm-router/src/catalog/resolver.rs
knjiang: yes but the proxy accepts format openai and will convert to google regardless

PR #170, opened 2026-04-01
fix(lingua): skip content ToolUse blocks when tool_calls is non-empty

evanmkeith:
## Summary

  LangGraph stores AIMessages in graph state with two representations of the
  same tool call: `content[].tool_use` (Anthropic/Bedrock provider format) and
  `tool_calls[]` (LangChain-native format). `parse_assistant_content` processed
  both without deduplication, emitting two `ToolCall` parts that rendered as
  duplicate entries in the LLM view tab.

  ## Change

  In `parse_assistant_content` (`langchain.rs`), skip `LangChainContentPartCompat::ToolUse`
  blocks in the `content[]` branch when `tool_calls` is non-empty. The
  `tool_calls[]` array is the canonical source in this case. Text parts in
  `content[]` continue to be processed normally.

  The early-return path (when `tool_calls.is_empty() == true`) is unchanged —
  providers that only populate `content[]` are unaffected.

  ## Notes

  - Visual-only bug. Raw span data in brainstore is unaffected.
  - Workaround for affected users: Thread view or JSON view.
  - Related: the surviving entry's arguments may still show a `$serde_json::private::Number` artifact for integer values due to a `serde_wasm_bindgen` / `serde_json::Value` serialization issue — tracked separately.

  ## Testing

  - [ ] Existing Lingua tests pass (`cargo test`)
  - [ ] New test: `parse_assistant_content` with dual-representation input produces exactly one `ToolCall` part
  - [ ] End-to-end: reproduction script at `pylon-13952/reproduce_tool_call_duplication.py` shows tool call once in LLM view tab

Linear bug [here](https://linear.app/braintrustdata/issue/BT-4608/lingua-langgraph-tool-calls-render-twice-in-llm-view-anthropicbedrock)

### DIFF
--- a/crates/lingua/src/processing/import/langchain.rs
+++ b/crates/lingua/src/processing/import/langchain.rs
@@ -408,16 +408,10 @@ fn parse_assistant_content(
                             provider_options: None,
                         }));
                     }
-                    LangChainContentPartCompat::ToolUse { id, name, input } => {
-                        parts.push(AssistantContentPart::ToolCall {
-                            tool_call_id: id,
-                            tool_name: name,
-                            arguments: parse_tool_call_arguments(Some(input)),
-                            encrypted_content: None,
-                            provider_options: None,
-                            provider_executed: None,
-                        });
-                    }
+                    // When tool_calls is non-empty, ToolUse blocks in content[] are the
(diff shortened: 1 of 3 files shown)


CLowbrow (approved):
please fix the fmt issues but lgtm

PR #174, opened 2026-04-02
use tool calls for json_object compatability with chat completions

knjiang:
This takes a similar approach to what we do in the proxy -> https://github.com/braintrustdata/braintrust-proxy/blob/main/packages/proxy/src/proxy.ts#L2791

i made the change and then ran `make regenerate-failed-transforms` which resulted in all the updated JSON files.

### DIFF
--- a/crates/lingua/src/providers/anthropic/convert.rs
+++ b/crates/lingua/src/providers/anthropic/convert.rs
@@ -1282,14 +1282,9 @@ impl TryFrom<&ResponseFormatConfig> for JsonOutputFormat {
     fn try_from(config: &ResponseFormatConfig) -> Result<Self, Self::Error> {
         match config.format_type.ok_or(())? {
             ResponseFormatType::Text => Err(()),
-            ResponseFormatType::JsonObject => Ok(JsonOutputFormat {
-                schema: serde_json::from_value(json!({
-                    "type": "object",
-                    "additionalProperties": false
-                }))
-                .expect("static JSON object is always a valid Map"),
-                json_output_format_type: JsonOutputFormatType::JsonSchema,
-            }),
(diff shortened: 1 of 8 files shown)


knjiang:
* **#174** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/lingua/src/providers/anthropic/convert.rs
```diff
@@ -1282,14 +1282,9 @@ impl TryFrom<&ResponseFormatConfig> for JsonOutputFormat {
...
         match config.format_type.ok_or(())? {
             ResponseFormatType::Text => Err(()),
-            ResponseFormatType::JsonObject => Ok(JsonOutputFormat {
-                schema: serde_json::from_value(json!({
-                    "type": "object",
-                    "additionalProperties": false
-                }))
-                .expect("static JSON object is always a valid Map"),
-                json_output_format_type: JsonOutputFormatType::JsonSchema,
-            }),
```
knjiang, line 1292: in practice:
```rust
JsonOutputFormat {
                schema: serde_json::from_value(json!({
                    "type": "object",
                    "additionalProperties": false
                }))
                .expect("static JSON object is always a valid Map"),
                json_output_format_type: JsonOutputFormatType::JsonSchema,
            }),
```
always emits an empty dictionary of `{}`, i thought we could set it like the above but no it doesn't work :( 

see the transforms JSON showing we now produce abritrary json

aswink (approved):

PR #179, opened 2026-04-07
add universal web search tool

knjiang:
This PR adds support for web search between all the providers and ensures regular + streaming calls look good.


knjiang:
* **#179** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/179?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/179?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

payloads/transforms/transform_errors.json
```diff
@@ -5,13 +5,11 @@
...
   },
   "anthropic_to_chat-completions": {
-    "webSearchToolParam": "Tool 'web_search' of type 'web_search_20250305' is not supported by openai",
     "codeInterpreterToolParam": "Tool 'bash' of type 'bash_20250124' is not supported by openai",
     "textEditorToolParam": "Tool 'str_replace_editor' of type 'text_editor_20250124' is not supported by openai",
     "textEditorToolV2Param": "Tool 'str_replace_based_edit_tool' of type 'text_editor_20250429' is not supported by openai",
     "textEditorToolV3Param": "Tool 'str_replace_based_edit_tool' of type 'text_editor_20250728' is not supported by openai",
-    "webSearchToolAdvancedParam": "Tool 'web_search' of type 'web_search_20250305' is not supported by openai",
-    "webSearchUserLocationParam": "Tool 'web_search' of type 'web_search_20250305' is not supported by openai",
+    "webSearchToolAdvancedParam": "Tool 'web_search' of type 'allowed_domains' is not supported by openai",
```
knjiang, line 12: `allowed_domains` shouldn't be translatable since there isn't an equivalent option for either Google or OpenAI.

i think this makes sense to error out instead of dropping the `allowed_domains` option since it feels like an important constraint defined by user that shouldn't be dropped.

crates/lingua/src/providers/anthropic/convert.rs
```diff
@@ -1373,6 +1374,32 @@ impl TryFrom<&UniversalTool> for Tool {
...
+                                field: format!("config for Google builtin tool '{}'", tool.name),
+                            })?;
+                    let _google_search: GoogleSearch =
+                        serde_json::from_value(config_val).map_err(|e| {
+                            ConvertError::JsonSerializationFailed {
+                                field: format!("Google search tool '{}'", tool.name),
+                                error: e.to_string(),
+                            }
+                        })?;
+                    return Ok(Tool::WebSearch20250305(generated::WebSearchTool20250305 {
```
aswink, line 1393: any reason not to use the newer one? web_search_20260209

aswink (approved):
maybe I missed it but are there also test cases from anthropic/openai -> google format?

crates/lingua/src/providers/anthropic/convert.rs
knjiang: gonna defer, but yes our generated types are out of date so we may need to regenerate.

https://github.com/braintrustdata/lingua/blob/main/crates/lingua/src/providers/anthropic/generated.rs#L947

PR #188, opened 2026-04-09
fix stream detection for google/bedrock which encode stream/non-streaming based on endpoint

knjiang:
for LLM providers such as Google/Bedrock, deciding streaming/non-streaming is codified in the endpoint you call. so we have to do some post-processing for clients that are calling from a Google endpoint and want streaming output.


knjiang:
* **#188** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/188?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/braintrust-llm-router/src/providers/mod.rs
```diff
@@ -140,6 +140,35 @@ pub(crate) fn disable_streaming_payload(payload: Bytes) -> Bytes {
...
+    };
+
+    match format {
+        ProviderFormat::ChatCompletions => {
+            object.insert("stream".into(), Value::Bool(true));
+        }
+        ProviderFormat::Responses | ProviderFormat::Anthropic => {
+            object.insert("stream".into(), Value::Bool(true));
+        }
+        ProviderFormat::Google
```
aswink, line 164: add a comment that these handle streaming via the path? it's somewhat confusing that this function does not enable streaming payloads in that case

crates/braintrust-llm-router/src/router.rs
```diff
@@ -113,16 +114,23 @@ async fn prepare_provider_request(
...
-        Err(err) => Err(err.into()),
+    let transformed = match lingua::transform_request(body.clone(), format, Some(&spec.model)) {
+        Ok(TransformResult::PassThrough(bytes)) => bytes,
+        Ok(TransformResult::Transformed { bytes, .. }) => bytes,
+        Err(TransformError::UnsupportedTargetFormat(_)) => body,
+        Err(err) => return Err(err.into()),
+    };
+
+    if stream {
+        Ok(enable_streaming_payload(transformed, format))
```
aswink, line 133: should this be part of transform_request?

aswink (approved):
wondering if there's a cleaner way to organize this, but lgtm otherwise

crates/braintrust-llm-router/src/router.rs
knjiang: i think we would have to add an extra boolean but i didnt want to go through the refactor endeavor yet. ill add a  TODO because i think having `transform_request` with a stream: boolean parameter seems more ergonomic.

PR #199, opened 2026-04-14
return metadata for logging when sending a completion request

aswink:
Split up the complete/complete_stream functions into create_request and complete.
The create_request returns some metadata that can be used for logging, regardless
of whether the provider request passes or fails. Include the detected input format,
provider alias, and provider format. Including this info here also removes the need
for the provider_aliases() function, so remove that. I never liked that clients used it
anyway because it's possible for it to diverge from what complete() actually does.

### DIFF
--- a/crates/braintrust-llm-router/src/router.rs
+++ b/crates/braintrust-llm-router/src/router.rs
@@ -266,9 +266,13 @@ impl Router {
         &self,
         model: &str,
         output_format: ProviderFormat,
-    ) -> Result<Vec<String>> {
-        self.resolve_providers(model, output_format)
-            .map(|routes| routes.into_iter().map(|(alias, ..)| alias).collect())
+    ) -> Result<Vec<(String, ProviderFormat)>> {
+        self.resolve_providers(model, output_format).map(|routes| {
+            routes
+                .into_iter()
+                .map(|(alias, _, _, _, format, _)| (alias, format))
+                .collect()
+        })
     }
 
     /// Resolve all providers for a given model and output format.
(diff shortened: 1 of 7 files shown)


aswink:
* **#199** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/199?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aswink:
actually I don't think this is necessary, let me close for now

aswink:
actually I think this would be helpful, reopening and including in https://github.com/braintrustdata/braintrust/pull/12794

crates/braintrust-llm-router/src/router.rs
```diff
@@ -154,123 +201,169 @@ impl Router {
...
-    /// # Arguments
-    ///
-    /// * `model` - The model name for routing (e.g., "gpt-4", "claude-3-opus")
-    /// * `output_format` - The output format, or None to auto-detect from body
-    ///
-    /// # Returns
-    /// A vector of provider aliases that can handle the given model and output
-    /// format. The aliases are in priority order. Follows the same order as the
-    /// complete and complete_stream methods.
-    pub fn provider_aliases(
```
knjiang, line 265: this code isn't necessary anymore?

knjiang (approved):

crates/braintrust-llm-router/src/router.rs
aswink: we could leave it in, but afaik it was only used for logging/tracing since there was no way to get this info from complete/complete_stream previously

knjiang: oh i see, yeah. we can rm

PR #200, opened 2026-04-17
anthropic: strip incompatible params for Opus 4.7

CLowbrow:
https://openrouter.ai/docs/guides/evaluate-and-optimize/model-migrations/claude-4-7#whats-new

### DIFF
--- a/crates/lingua/src/providers/anthropic/capabilities.rs
+++ b/crates/lingua/src/providers/anthropic/capabilities.rs
@@ -27,8 +27,8 @@ fn part_supports_output_config_effort(model_part: &str) -> bool {
 /// Transforms required for specific Anthropic model families.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ModelTransform {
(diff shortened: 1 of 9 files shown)


CLowbrow:
Some extra stuff in here to make CI happy while I fix it

crates/lingua/src/providers/anthropic/capabilities.rs
```diff
@@ -23,10 +24,71 @@ fn part_supports_output_config_effort(model_part: &str) -> bool {
...
+            return Some(transforms);
+        }
+    }
+    None
+}
+
+/// Get the transforms required for a model.
+pub fn get_model_transforms(model: &str) -> &'static [ModelTransform] {
+    let lower = model.to_ascii_lowercase();
+    // Bedrock/Vertex model IDs wrap the Anthropic model token with provider-specific
```
aswink, line 53: would be nice if we eventually made the whole "model ID" vs "model name" simpler to deal with, but that's a problem for another day

aswink (approved):

PR #204, opened 2026-04-27
[bug] google to chat completions and google to responses

cpinn:
Made this small change to fix a parsing error. I believe the right location to fix this is in tools but I also noticed some changes in adapters and it is unclear to me where this change should appropriately live.

For google to open ai responses set code_interpreter property for codeExecution google property.
For google to open ai chat completions strip out the code_interpreter property as there appears to be no equivalent.

Fix the existing tests and remove the transform errors.


CLowbrow (approved):
Did you test this by sending requests through gateway with these tools and seeing that the functionality is the same? I'm worried about transforming tools provided by providers.

PR #205, opened 2026-04-28
BT-4991 Fix transform when vertex returns a complete anthropic response

erin2722:
Vertex will return complete responses from Anthropic, even when the streaming format is specified:
```
{
  "type": "message",
  "content": [...]
}
```
This causes us to be unable to parse the output, which leads to a silent failure and a "-" rendered in the UI given that this is an empty output. My understanding is that we want to keep the silent failure behavior, as sometimes the format is able to just be passed straight through, and so instead I've added info-level logging for these events in addition to fixing this at the transform layer.

This change converts the response format into an expected streaming-formatted resposne:
```
{
  "choices": [{
    "delta": {
      "content": "..."
    }
  }]
}
```

Before fix (current behavior in prod):
Screenshot 2026-04-28 at 5 15 28 PM

After fix (localhost):
Screenshot 2026-04-28 at 5 15 41 PM


crates/lingua/src/processing/transform.rs
```diff
@@ -377,7 +438,31 @@ pub(crate) fn transform_stream_chunk_step(
         .map_err(|e| TransformError::DeserializationFailed(e.to_string()))?;
     let event_type = extract_event_type(&chunk);
 
-    let source_adapter = detect_adapter(&chunk, DetectKind::Stream)?;
+    let source_adapter = match detect_adapter(&chunk, DetectKind::Stream) {
+        Ok(adapter) => adapter,
+        Err(TransformError::UnableToDetectFormat) => {
```
erin2722: erin todo from offline convo: move this logic into detect_adapter

payloads/cases/params.ts
```diff
@@ -97,8 +97,7 @@ export const paramsCases: TestCaseCollection = {
             },
             {
               type: "input_file",
-              file_url:
-                "https://www.berkshirehathaway.com/letters/2024ltr.pdf",
+              file_url: "https://www.berkshirehathaway.com/letters/2024ltr.pdf",
```
erin2722, line 100: this was automatically linted w my pre-commit hooks

crates/braintrust-llm-router/src/streaming.rs
```diff
@@ -10,6 +10,34 @@ use reqwest::Response;
...
+                let payload_keys = value
+                    .as_object()
+                    .map(|obj| obj.keys().cloned().collect::<Vec<_>>().join(","))
+                    .unwrap_or_default();
+                (payload_type, payload_keys)
+            })
+            .unwrap_or_default();
+
+    tracing::info!(
+        router.stream.input_bytes = data.len(),
```
erin2722, line 33: Should we avoid logging input here? Is this PII?

crates/lingua/src/processing/stream.rs
```diff
@@ -397,11 +405,29 @@ fn expand_anthropic_session_chunks(
...
+        let content = delta_view
             .as_ref()
             .and_then(|d| d.content.as_deref())
-            .filter(|s| !s.is_empty())
-        {
+            .filter(|s| !s.is_empty());
+        let is_full_anthropic_response =
+            source_format == ProviderFormat::Anthropic && !source_is_native_stream;
+
+        if is_full_anthropic_response && !anthropic_message_started {
```
erin2722, line 428: `source_is_native_stream` tells us whether or not the response is already properly formatted as a streaming response.

The below logic ensures that we re-write the "full" / non-stream formated vertex-anthropic streaming response into a streaming format that is Anthropic-compatible, specifically by adding the message_start, content_block_start, and content_block_stop fields.

Comparing:
https://github.com/braintrustdata/lingua/blob/d17a9512e97a8d2aa5839f33a12de36540bdb9b9/payloads/snapshots/simpleRequest/anthropic/response-streaming.json#L4

and
https://github.com/braintrustdata/lingua/blob/d17a9512e97a8d2aa5839f33a12de36540bdb9b9/payloads/snapshots/simpleRequest/vertex-anthropic/response-streaming.json

Helps visualize this

payloads/cases/simple.ts
```diff
@@ -270,6 +270,30 @@ export const simpleCases: TestCaseCollection = {
     },
   },
 
+  thinkingSignatureRequest: {
```
erin2722, line 273: I added this test case to ensure that the thinking blocks are properly re-written as chunks from the "full" vertex responses

payloads/scripts/transforms/helpers.ts
```diff
@@ -361,6 +366,12 @@ export const STREAMING_PAIRS: TransformPair[] = [
     wasmSource: "Google",
     wasmTarget: "OpenAI",
   },
+  {
+    source: "chat-completions",
+    target: "vertex-anthropic",
+    wasmSource: "OpenAI",
+    wasmTarget: "Anthropic",
```
erin2722, line 373: My understanding is that `"Anthropic"` is still the right wasmTarget here, because `VertexAnthropic` is internal-only and probably shouldn't yet be exposed publicly from lingua. Ie, the guarantee here should be that "anthropic responses are properly formatted as anthropic responses, regardless of whether they come from vertex or anthropic itself". LMK if that seems off

crates/lingua/src/processing/transform.rs
```diff
@@ -369,6 +374,89 @@ pub(crate) fn serialize_stream_value(value: &Value) -> Result<Bytes, TransformEr
     )?))
 }
 
+fn assistant_content_to_stream_delta(content: &AssistantContent) -> UniversalStreamDelta {
```
erin2722, line 381: This function re-writes the individual "full" response into a chat-completion compatible stream chunk:
```
    "content": [
      {
        "type": "text",
        "text": "The capital of France is Paris."
      }
    ],
```

into 

```
    "choices": [
      {
        "index": 0,
        "delta": {
          "role": "assistant",
          "content": "The capital of France is Paris",
          "refusal": null
        },
        "finish_reason": null
      }
    ],
```

crates/braintrust-llm-router/src/streaming.rs
CLowbrow: length of bytes? prooobably not? I'm open to the idea that you could learn something from the length of the message but I'm not super worried about it?

CLowbrow (approved):
nice

erin2722:
looking at a CI failure: https://github.com/braintrustdata/braintrust/actions/runs/25172999113/job/73797578295?pr=13469#step:7:7979

before merging

PR #207, opened 2026-04-29
Make mime type stuff work

CLowbrow:
Vetex was having issues without mime types

shamelessly copying https://github.com/braintrustdata/lingua/pull/206

### DIFF
--- a/crates/lingua/src/providers/google/convert.rs
+++ b/crates/lingua/src/providers/google/convert.rs
@@ -68,20 +68,48 @@ fn mime_type_from_url(url: &str) -> String {
                 Some("image/jpeg")
             } else if extension.eq_ignore_ascii_case("png") {
                 Some("image/png")
-            } else if extension.eq_ignore_ascii_case("gif") {
-                Some("image/gif")
             } else if extension.eq_ignore_ascii_case("webp") {
                 Some("image/webp")
-            } else if extension.eq_ignore_ascii_case("bmp") {
-                Some("image/bmp")
-            } else if extension.eq_ignore_ascii_case("svg") {
-                Some("image/svg+xml")
+            } else if extension.eq_ignore_ascii_case("heic") {
+                Some("image/heic")
+            } else if extension.eq_ignore_ascii_case("heif") {
+                Some("image/heif")
             } else if extension.eq_ignore_ascii_case("pdf") {
                 Some("application/pdf")
             } else if extension.eq_ignore_ascii_case("txt") {
                 Some("text/plain")
-            } else if extension.eq_ignore_ascii_case("json") {
-                Some("application/json")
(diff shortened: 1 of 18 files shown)


crates/lingua/src/providers/google/convert.rs
```diff
@@ -55,6 +55,46 @@ fn text_part(text: String) -> GooglePart {
     }
 }
 
+fn mime_type_from_url(url: &str) -> String {
```
jessicaalder-lgtm, line 58: Is this worth generalizing as a public helper in case it is needed outside of Google?

crates/lingua/src/providers/google/convert.rs
```diff
@@ -55,6 +55,46 @@ fn text_part(text: String) -> GooglePart {
...
+    if let Some(metadata) = parse_file_metadata_from_url(url) {
+        if let Some(content_type) = metadata.content_type {
+            return content_type;
+        }
+
+        if let Some((_, extension)) = metadata.filename.rsplit_once('.') {
```
jessicaalder-lgtm, line 64: May be worth covering audio (mp3/wav/m4a) and video (mp4/mov/webm) for Gemini?

crates/lingua/src/providers/google/convert.rs
```diff
@@ -55,6 +55,46 @@ fn text_part(text: String) -> GooglePart {
...
+            } else if extension.eq_ignore_ascii_case("png") {
+                Some("image/png")
+            } else if extension.eq_ignore_ascii_case("gif") {
+                Some("image/gif")
+            } else if extension.eq_ignore_ascii_case("webp") {
+                Some("image/webp")
+            } else if extension.eq_ignore_ascii_case("bmp") {
+                Some("image/bmp")
+            } else if extension.eq_ignore_ascii_case("svg") {
+                Some("image/svg+xml")
```
jessicaalder-lgtm: Not sure how meaningful, but Vertex Gemini models do not support svg/xml or bmp: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-flash

jessicaalder-lgtm (approved):
Looks good to me, nice adding the cross-provider expected-differences and snapshots too. I added some comments, but nothing blocking if you think they're fine. LGTM

crates/lingua/src/providers/google/convert.rs
CLowbrow: Ehhh im not sure vertex has weird requirements around what mime types it accepts I'd like to keep it separate.

crates/lingua/src/providers/google/convert.rs
CLowbrow: good catch i'll use the actual vertex list.

PR #214, opened 2026-05-01
more correct reasoning effort planning for oai

CLowbrow:
trying to clamp per-model when transforming to openai requests from something else.

### DIFF
--- a/crates/lingua/src/universal/reasoning.rs
+++ b/crates/lingua/src/universal/reasoning.rs
@@ -185,23 +185,39 @@ impl From<&Thinking> for ReasoningConfig {
     }
 }
 
-fn from_generated_openai_effort(effort: &GeneratedOpenAIReasoningEffort) -> ReasoningEffort {
-    match effort {
-        GeneratedOpenAIReasoningEffort::Minimal => ReasoningEffort::Minimal,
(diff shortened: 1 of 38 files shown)


crates/lingua/src/universal/reasoning.rs
```diff
@@ -176,17 +185,33 @@ impl From<&Thinking> for ReasoningConfig {
...
+fn from_generated_openai_effort(effort: &GeneratedOpenAIReasoningEffort) -> ReasoningEffort {
+    match effort {
+        GeneratedOpenAIReasoningEffort::Minimal => ReasoningEffort::Minimal,
+        GeneratedOpenAIReasoningEffort::Low => ReasoningEffort::Low,
+        GeneratedOpenAIReasoningEffort::Medium => ReasoningEffort::Medium,
+        GeneratedOpenAIReasoningEffort::High => ReasoningEffort::High,
+    }
+}
+
+fn from_openai_effort_param(effort: OpenAIReasoningEffortParam) -> ReasoningEffort {
```
knjiang: can we use the rust from trait from these? https://doc.rust-lang.org/std/convert/trait.From.html

knjiang (approved):

PR #215, opened 2026-05-04
Fix Gemini tool calling in v2 and v2.5 models

erin2722:
This fix addresses two issues:

- The finish_reason for chat completion tool calls can validly be stop, but to gemini this will stop it from looking to make tool calls. We need to translate the finish_reason into tool_calls
- Each tool call requires a unique id to be associated with it, and so we attach that here as well.


erin2722:
Testing here is a bit weird/difficult, and I'd like some feedback here:

The `payloads/` tests have tool call testing, but because the test harness generates the tool-call results itself ([here](https://github.com/braintrustdata/lingua/blob/87456835fc318d6f575a514b130b65e5e60062dd/payloads/scripts/providers/google.ts#L200-L216)), we don't actually see that the initial `finishReason` [in the response capture](https://github.com/braintrustdata/lingua/blob/87456835fc318d6f575a514b130b65e5e60062dd/payloads/snapshots/toolCallRequest/google/response-streaming.json#L19) would've caused gemini to stop issuing follow-up requests.

Similarly, we don't see that the missing ids would've caused followup tool calls to fail.

I modified [verify_proxy_models.ts](https://github.com/braintrustdata/braintrust-proxy/blob/6f22bbf096f931440d9d761716350b75c32e2950/packages/proxy/scripts/verify_proxy_models.ts#L4) to run  against the gateway endpoint and use a tool_call request, and used that for my local iteration loop when testing gemini. I wonder if we would want to pull out this test infra from the proxy and share it with the gateway, and then run the set of test cases in payloads/cases/ with real provider calls throughout the conversation. This might make local iteration easier, and if we detected specific lingua changes in PRs, we could run those specific provider tests as part of CI. 

I _think_ this would fill a testing gap of "can the providers actually process the requests and responses that lingua produces?", but I'm still learning the test infra and would love some feedback on this observation/idea.


---

Note: After talking with Ken, I understand that the existing payloads/ tests are moreso meant to be regression tests, rather than proactively catching these issues. I've just introduced the chat-completions=>google streaming pair to payloads/, which should be sufficient for this change

knjiang (approved):

PR #216, opened 2026-05-04
account for conversation reference id for responses

CLowbrow:
Alternative to this is https://github.com/braintrustdata/lingua/pull/212 where we just drop this.


knjiang (approved):

cpinn (approved):
Looks good to me. 

I think likely we may need to evolve this depending on what other providers are doing. I was looking into openai a bit more and it seems like different types of requests rs_ v msg_ v item_ can be stuck into the id format. I think this should work for our use case and is better than stripping it out and I think it will just evolve as we learn more.

PR #226, opened 2026-05-07
upgrade cargo dependencies

cpinn:
Fix a number of critical and high vulnerabilities in aws-lc-sys, openssl, rustls-webpki. 

In order to fix rustls-webpki needed to upgrade rust to 1.91.1 in order to pull in a later version of aws sdk bedrock.


cpinn:
folded into other changes

PR #231, opened 2026-05-08
don't need convert map to objects anymore

j13huang:
According to claude, we don't need the `convertMapsToObjects` function anymore because we serialize everything directly into json now?

https://github.com/braintrustdata/lingua/pull/191

So in theory we could do away with this extra overhead. Please tell me if this doesn't make sense

### DIFF
--- a/bindings/typescript/src/converters.ts
+++ b/bindings/typescript/src/converters.ts
@@ -53,35 +53,6 @@ export class ConversionError extends Error {
 // ============================================================================
 
 /**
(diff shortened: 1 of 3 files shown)


choochootrain (commented):
tagging in @knjiang   
  
seems fine but we should make sure we dont break topics or gateway

knjiang:
ya seems fine

knjiang (approved):

PR #233, opened 2026-05-11
Auto-route /chat/completions to /responses for unsupported chat completion features

erin2722:
This fix addresses an issue where the `/chat/completions` endpoint doesn't support reasoning with tool calls, as part of OpenAI's push to get every to migrate to the `/responses` endpoint. Lingua was passing these requests straight through to the configured `/chat/completions` endpoint and we were encountering a failure because of this.

Like we were chatting about in standup last week, there are 2 potential resolutions to this:

- Fix the UI to let people delete the reasoning fields, and then return a clearer error message explaining to unset the reasoning field here
- Automatically forward these requests to `/responses` based on additional specific format detection logic in the router.

I've gone with the second option here-- from my reading, I'd expect us to run into this flavor of problem again as openai pushes everyone to responses. Its not _too_ janky :), and I think will come in handy in the future too.

### DIFF
--- a/crates/lingua/src/providers/openai/convert.rs
+++ b/crates/lingua/src/providers/openai/convert.rs
@@ -2333,8 +2333,13 @@ impl TryFromLLM<Vec<openai::OutputItem>> for Vec<Message> {
 
             let parts: Vec<AssistantContentPart> = match item.output_item_type {
                 Some(openai::OutputItemType::Message) => {
-                    // Extract text content from message output items
+                    // Extract text content from message output items.
+                    // `phase` is a message-level field (not content-level); it is stored in
+                    // the first text part's provider_options under "_output_item_phase" so
+                    // it survives the OutputItem → Message → OutputItem roundtrip.
+                    let item_phase = item.phase.take();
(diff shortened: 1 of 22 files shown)


crates/lingua/src/providers/openai/convert.rs
```diff
@@ -2333,8 +2333,13 @@ impl TryFromLLM<Vec<openai::OutputItem>> for Vec<Message> {
 
             let parts: Vec<AssistantContentPart> = match item.output_item_type {
                 Some(openai::OutputItemType::Message) => {
-                    // Extract text content from message output items
+                    // Extract text content from message output items.
+                    // `phase` is a message-level field (not content-level); it is stored in
```
erin2722, line 2337: This was a bit of a seperate change that I found because of the additional payload tests. We were stripping the `phase` parameter from messages when going through the universal format, and while this doesn't cause an error, it can negatively impact the chat responses-- https://developers.openai.com/api/docs/guides/reasoning#phase-parameter

These changes stick it in provider_options for the universal round trip

knjiang (approved):

PR #237, opened 2026-05-12
allow passing in of custom client settings

ibolmo:


crates/braintrust-llm-router/src/client.rs
```diff
@@ -20,13 +21,21 @@ const DEFAULT_MAX_RETRIES: u32 = 2;
 // Shared reqwest clients are cached by these settings. Keep this key
```
aswink: keep this comment next to ClientSettings?

crates/braintrust-llm-router/src/client.rs
```diff
@@ -77,19 +96,25 @@ pub fn build_middleware_client(settings: &ClientSettings) -> Result<ClientWithMi
 }
 
 fn build_middleware_client_inner(settings: &ClientSettings) -> Result<ClientWithMiddleware> {
+    let cacheable = settings.dns_overrides.is_empty();
```
aswink: why not cache if we have overrides? seems like it could have some perf impact

aswink (approved):

PR #240, opened 2026-05-18
Propagate custom headers through llm-router

erin2722:


knjiang (commented):
can we do this mostly in main repo? have it apply into client_headers?

erin2722:
I initially avoided doing this because we filter out some specific headers for these [ClientHeaders here](https://github.com/braintrustdata/braintrust/blob/9545e33ad7fd83effad32bb41e267919d02c830c/gateway/src/proxy/types.rs#L24-L75) and [here](https://github.com/braintrustdata/lingua/blob/e57b3b524ce2d2e6426877b1db3f78ba473cb868/crates/braintrust-llm-router/src/providers/mod.rs#L36-L60), and its unclear to me that we need to apply these same filters to the configured request headers here, esp bc our [docs](https://www.braintrust.dev/docs/integrations/ai-providers/custom#custom-headers-and-templating) call out authorization as a header that may be configured by a user.

We could add a new field to `ClientHeaders` that doesn't get filtered, which might be a bit cleaner, or I might be off-base here that we don't want to apply the filters to these configured headers, and we can just make these client headers.

WDYT?

---

after writing it out I think at the very least I'll want to move this into client headers, but curious abt your thoughts as to whether we need to apply the filtering to these headers

knjiang:
ya client headers unfiltered sgtm

knjiang (approved):

PR #244, opened 2026-05-19
upgrade npm dependencies

cpinn:


payloads/pnpm-lock.yaml
```diff
@@ -1,1940 +0,0 @@
-lockfileVersion: "9.0"
```
cpinn, line 1: this is old and is handled by the workspace, checked out a new repo and ran commands on the payloads.

CLowbrow (approved):

PR #258, opened 2026-05-28
Fixes for Opus 4.8

erin2722:
We were hardcoding opus 4.7 in a few places, modify to use a regex to catch opus 4.8 and beyond


knjiang (approved):

erin2722:
@codex review

PR #259, opened 2026-05-28
Fix opus regex

erin2722:


crates/lingua/src/providers/anthropic/adapter.rs
```diff
@@ -1852,6 +1852,7 @@ mod tests {
         let adapter = AnthropicAdapter;
 
         for model in [
+            "claude-opus-4-20250514",
```
erin2722, line 1855: I didn't realize this was a valid model name 😢 fixed

knjiang (approved):

PR #264, opened 2026-06-01
enable raw response logging

CLowbrow:
Want to be able to log untransformed responses


crates/braintrust-llm-router/src/streaming.rs
aswink, line 74: I wonder if another possible version is to return a stream that returns something like `StreamChunkWithRaw`, instead of adding a callback function? Just so we get all results the same way and don't have to worry about blocking this loop with the callback function. I don't feel too strongly about this, though

crates/braintrust-llm-router/src/router.rs
```diff
@@ -305,6 +309,25 @@ impl Router {
...
+    ) -> Result<Bytes> {
+        self.complete_internal(request, client_headers, None).await
+    }
+
+    pub async fn complete_with_raw_response_capture(
+        &self,
+        request: PreparedRequest<'_>,
+        client_headers: &ClientHeaders,
+        raw_response_capture: RawResponseCapture,
+    ) -> Result<Bytes> {
```
aswink: If we change the stream to return both results together, this could also do the same and return a pair or a struct here, of two `Bytes` objects

crates/braintrust-llm-router/src/streaming.rs
CLowbrow: I think this is right but is also hard because raw chunks can produce 0+ transformed chunks. I am cowardly going to back out of trying this for now.

aswink: That's fair. Maybe add a comment that callback does block the response stream so not to do anything too crazy in it?

aswink (approved):

PR #271, opened 2026-06-04
let the google-flavor chat completions through

CLowbrow:
Test:

```bash
curl -i http://localhost:8080/proxy/v1/models/gemini-2.5-flash:generateContent \
  -H "Authorization: Bearer ${BRAINTRUST_API_KEY}" \
  -d '{
    "contents": [
      {
        "parts": [
          {"text": "What is one recent AI news item? Keep it short."}
        ]
      }
    ],
    "tools": [
      {
        "google_search": {}
      }
    ]
  }'
```

### DIFF
--- a/crates/braintrust-llm-router/src/providers/google.rs
+++ b/crates/braintrust-llm-router/src/providers/google.rs
@@ -9,7 +9,7 @@ use crate::streaming::{sse_stream, RawResponseStream};
 use async_trait::async_trait;
 use bytes::Bytes;
 use lingua::ProviderFormat;
-use reqwest::header::HeaderMap;
+use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION};
 use reqwest::Url;
 use reqwest_middleware::ClientWithMiddleware;
 
@@ -69,6 +69,13 @@ impl GoogleProvider {
         Self::new_with_client_settings(config, client_settings.unwrap_or_default())
     }
 
+    fn chat_completions_url(&self) -> Result<Url> {
+        self.config
+            .endpoint
+            .join("openai/chat/completions")
+            .map_err(|e| Error::InvalidRequest(format!("invalid Google endpoint: {e}")))
+    }
+
     fn generate_url(&self, model: &str, stream: bool) -> Result<Url> {
         let path = if model.starts_with("models/") {
             model.to_string()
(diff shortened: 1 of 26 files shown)


crates/lingua/src/processing/transform.rs
```diff
@@ -273,6 +273,33 @@ pub fn transform_request(
...
+    transform_request_with_options(
+        input,
+        target_format,
+        model,
+        TransformRequestOptions::default(),
+    )
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct TransformRequestOptions {
```
knjiang: instead of introducing a new abstraction here, can we resolve this in resolver.rs? https://github.com/braintrustdata/lingua/blob/main/crates/braintrust-llm-router/src/catalog/resolver.rs#L42

knjiang: here's what my AI said:
```
Responses only came up because the current diff added TransformRequestOptions to stop this case:

  Google provider selected as ChatCompletions
  → lingua::transform_request(...) sees reasoning_effort + tools
  → it auto-upgrades target format to Responses
  → router would send Google traffic to an OpenAI Responses-shaped route, which is wrong for Google.

  But you’re right: if the goal is just Google routing, we should not solve it with a new transform abstraction. The
  cleaner plan is:

  - Keep Google’s provider_formats() as [Google, ChatCompletions].
  - Move the Google-specific routing decision into catalog/resolver.rs, alongside the existing Bedrock/Vertex model-name
    routing rules.
  - Remove TransformRequestOptions and call plain lingua::transform_request(...).
  - Make sure Gemini API models like gemini-2.5-flash resolve to alias google, and when requested as ChatCompletions,
    route to Google’s OpenAI-compatible /openai/chat/completions endpoint.
  - Keep Vertex models like publishers/google/models/... routed through Vertex/native Google format.

  So Responses is not the feature; it’s just the bug the current abstraction was trying to avoid. The fix should be Google
  routing in resolver/provider-format selection.
```

crates/braintrust-llm-router/src/providers/google.rs
```diff
@@ -69,6 +69,26 @@ impl GoogleProvider {
...
+                .path_segments_mut()
+                .map_err(|_| Error::InvalidRequest("endpoint must be absolute".into()))?;
+            segments.pop_if_empty();
+            if !has_openai_suffix {
+                segments.push("openai");
+            }
+            segments.push("chat");
+            segments.push("completions");
+        }
+        Ok(url)
```
knjiang, line 89: is this complcated? i feel like constructing a url should be easier

knjiang (approved):

crates/braintrust-llm-router/src/providers/google.rs
CLowbrow: I mean it's verbose, but I don't want to try to parse URLs myself. Easier how?

knjiang: oh i see because it can be configurable, ignore me

PR #277, opened 2026-06-08
Add `x-anthropic-prompt-cache` to enable prompt caching for anthropic models

knjiang:
when using other apis such as responses/gemini, you can't enable prompt caching for anthropic models so adding a header as a workaround.


knjiang:
* **#277** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/277?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/277?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

erin2722 (approved):
LGTM mod the codex comment which is correct

it makes me a little uneasy that we are introducing this new dimension of custom header-based behavior to lingua, but I don't think there is another way -- if this ends up becoming a common case, I think we will want to refactor this out of the provider-specific logic.

PR #283, opened 2026-06-09
Separate route planning from request execution in the llm router

erin2722:


crates/braintrust-llm-router/src/router.rs
```diff
@@ -130,14 +130,25 @@ pub fn create_provider(
...
+
+impl ProviderRoute {
+    pub fn provider_alias(&self) -> &str {
+        self.provider_alias.as_str()
+    }
+
+    pub fn model(&self) -> &str {
+        self.spec.model.as_str()
+    }
+}
```
knjiang, line 150: why did this have to become a struct/ add new methods?

crates/braintrust-llm-router/src/router.rs
```diff
@@ -158,19 +169,36 @@ pub struct RouterMetadata {
     pub provider_format: ProviderFormat,
 }
 
+/// Provider route resolved for a caller-specified provider alias.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ResolvedProviderRoute {
+    pub provider_alias: String,
+    pub wire_format: ProviderFormat,
+}
```
knjiang: same here, i guess its for the field names?

knjiang (approved):

crates/braintrust-llm-router/src/failover.rs
```diff
@@ -0,0 +1,189 @@
...
+    catalog: &ModelCatalog,
+    model: &str,
+    candidates: Vec<FailoverRouteCandidate>,
+) -> Option<FailoverRoutePlan> {
+    let routes = candidates
+        .into_iter()
+        .filter(|candidate| {
+            candidate.is_configured
+                && (candidate.custom_model_supported
+                    || model_supports_any_provider_alias(
```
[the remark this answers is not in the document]
erin2722: will fix this in a followup to add custom model support

crates/braintrust-llm-router/src/router.rs
erin2722: So that it could be publicly accessible from the gateway-- now, lingua returns a list of routes (previously it was just internally using the first route), and then the gateway determines which route to pass into `create_request`, to let gateway handle the route planning and failover (and future route desicions as well)

erin2722: And the provider_alias / model functions are available mostly so that the gateway can have clear observability into the provider and model failover attempts (well there isn't model failover yet, but its still nice to just expose this info from the route)

crates/braintrust-llm-router/src/router.rs
erin2722: this one can be removed actually

PR #286, opened 2026-06-09
accept openai legacy prompts

CLowbrow:
Legacy prompt compatibility now detects valid completions-era prompt shapes even when they contain parameters whose types differ from Chat Completions, and rejects unsupported options like numeric logprobs with a clear compatibility error.

## Test

```bash
curl -sS -i https://api.braintrust.dev/v1/proxy/chat/completions \

  -H "Authorization: Bearer $BRAINTRUST_API_KEY" \

  -d '{

    "model": "gpt-4o",

    "prompt": "Write a haiku about the ocean.",

    "max_tokens": 256,

    "temperature": 0.7,

    "stream": false

  }'
```

### DIFF
--- a/crates/lingua/src/processing/transform.rs
+++ b/crates/lingua/src/processing/transform.rs
@@ -934,6 +934,24 @@ mod tests {
 
     #[test]
     #[cfg(feature = "openai")]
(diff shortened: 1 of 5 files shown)


knjiang (approved):
can we add a integration test to gateway when bumping

PR #291, opened 2026-06-10
Preserve mixed Anthropic tool results in cross-provider transforms

knjiang:
### Problem

Anthropic Messages allows a user message like:

```json
[
  { "type": "text", "text": "Use this output..." },
  { "type": "tool_result", "tool_use_id": "call_123", "content": "..." }
]

Lingua previously dropped the tool_result from mixed user content. That made OpenAI transforms invalid and made Google
transforms lose important context.
```

### Behavior change

This fixes mixed Anthropic user content where a message contains both text and a `tool_result`.

Before this change, Anthropic import dropped the `tool_result` from mixed user content blocks.

For OpenAI Responses, that produced an invalid request:

```json
[
  { "role": "user", "content": "Look up the latest records." },
  { "type": "function_call", "call_id": "call_repro_123" },
  { "role": "user", "content": [{ "type": "input_text", "text": "What details are available?" }] }
]
```

OpenAI rejected this with:

400 No tool output found for function call call_repro_123.

After this change, the transformed Responses request includes the required tool output:
```json
{
  "type": "function_call_output",
  "call_id": "call_repro_123",
  "output": "{\"records\":[{\"id\":\"record_1\",\"status\":\"ok\"}]}"
}
```

Chat Completions had the same protocol issue: an assistant tool_calls message was followed by a user message instead of a
matching role: "tool" message. The fixed transform now emits the required tool message.

For Google/Gemini, the request was accepted before, but the semantics were degraded because the function response was
dropped. Before, Gemini saw the function call and the follow-up user text, but not the actual tool result. After this
change, Gemini receives the preserved functionResponse:

```json
{
  "role": "user",
  "parts": [
    {
      "functionResponse": {
        "name": "search_records",
        "response": {
          "records": [{ "id": "record_1", "status": "ok" }]
        }
      }
    }
  ]
}
```

So this change should fix the translation for both when going from anthropic -> google / anthropic -> responses / anthropic -> chat completions

### DIFF
--- a/crates/lingua/src/providers/anthropic/adapter.rs
+++ b/crates/lingua/src/providers/anthropic/adapter.rs
@@ -15,7 +15,9 @@ use crate::processing::adapters::{
 };
 use crate::processing::transform::TransformError;
 use crate::providers::anthropic::capabilities;
-use crate::providers::anthropic::convert::system_to_user_content;
+use crate::providers::anthropic::convert::{
+    anthropic_input_messages_to_universal_messages, system_to_user_content,
+};
 use crate::providers::anthropic::detect::{
     system_messages_are_supported_and_well_placed, try_parse_anthropic_source,
 };
@@ -215,7 +217,7 @@ impl ProviderAdapter for AnthropicAdapter {
             TransformError::ToUniversalFailed("Anthropic: missing 'messages' field".to_string())
         })?;
 
-        let mut messages = <Vec<Message> as TryFromLLM<Vec<_>>>::try_from(input_messages)
+        let mut messages = anthropic_input_messages_to_universal_messages(input_messages)
             .map_err(|e| TransformError::ToUniversalFailed(e.to_string()))?;
 
         if let Some(system) = typed_params.system.clone() {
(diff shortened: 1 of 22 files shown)


knjiang:
* **#291** <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/braintrustdata/lingua/291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/lingua/src/providers/anthropic/adapter.rs
```diff
@@ -215,7 +218,7 @@ impl ProviderAdapter for AnthropicAdapter {
             TransformError::ToUniversalFailed("Anthropic: missing 'messages' field".to_string())
         })?;
 
-        let mut messages = <Vec<Message> as TryFromLLM<Vec<_>>>::try_from(input_messages)
+        let mut messages = anthropic_input_messages_to_universal_messages(input_messages)
```
knjiang, line 221: we can't use TryFromLLM because one anthropic input message can equal many universal messages so it breaks the 1:1 semantics that all the other adapters/conversions currently hold

knjiang:
@codex review

knjiang:
@codex review

CLowbrow (approved):

PR #308, opened 2026-06-15
Fix google thinking transform

CLowbrow:
## Why is this diff so big?

Google can put reasoning, visible text, signatures, tool starts, and finish data in the same provider chunk, while Responses and Anthropic require separate ordered events.

## What

Fix Google includeThoughts / reasoning stream handling so thought parts are kept separate from visible output instead of being flattened into parse-breaking content.

## Changes

- Preserve Google `thought: true` parts as universal reasoning deltas instead of assistant content.
- Preserve Google `thoughtSignature` values through reasoning, text, and function-call stream chunks.
- Emit valid target stream event sequences for mixed reasoning + text/tool chunks:
  - Google -> Responses emits separate reasoning, text, tool, and terminal events.
  - Google -> Anthropic emits ordered thinking/signature/text/tool blocks with valid content block indexes.
  - Google -> Chat Completions keeps visible content separate from reasoning.
- Normalize Responses stream output item indexes:
  - Responses `output_index` is treated as an output item index, not a universal choice index.
  - Responses function-call output item indexes are remapped to tool-relative indexes in stateful stream transforms.
- Preserve Google-native `thinkingBudget` roundtrips without forcing Anthropic/Bedrock budget reasoning into Gemini 3 `thinkingBudget`.
- Add recorded payload coverage for `googleThinkingJsonSchemaParam`.
- Update transform snapshots and narrow expected-diff entries for the intentional remaining provider differences.

### DIFF
--- a/crates/lingua/src/providers/google/adapter.rs
+++ b/crates/lingua/src/providers/google/adapter.rs
@@ -95,11 +95,16 @@ impl ProviderAdapter for GoogleAdapter {
                     let budget_tokens = tc.thinking_budget;
                     let effort = budget_tokens
                         .map(|b| crate::universal::reasoning::budget_to_effort(b, max_tokens));
+                    let canonical = if tc.thinking_budget.is_some() {
+                        Some(crate::universal::ReasoningCanonical::GoogleThinkingBudget)
+                    } else {
+                        Some(crate::universal::ReasoningCanonical::GoogleIncludeThoughts)
+                    };
                     crate::universal::ReasoningConfig {
                         enabled: Some(!is_disabled),
                         effort,
                         budget_tokens,
-                        canonical: Some(crate::universal::ReasoningCanonical::GoogleThinkingBudget),
+                        canonical,
                         ..Default::default()
                     }
                 }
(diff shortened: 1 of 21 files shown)


payloads/scripts/transforms/__snapshots__/transforms.test.ts.snap
```diff
@@ -20918,6 +20918,91 @@ exports[`google → anthropic > googleResponseSchemaPropertyOrderingParam > resp
 }
 `;
 
+exports[`google → anthropic > googleThinkingJsonSchemaParam > request 1`] = `
```
knjiang, line 20921: we didn't have to regenerate any existing stuff? does this mean we weren't really testing this sort of chunks before?

just curiosu on if we were just fully missing coverage.

crates/lingua/src/providers/google/adapter.rs
```diff
@@ -95,11 +95,16 @@ impl ProviderAdapter for GoogleAdapter {
                     let budget_tokens = tc.thinking_budget;
                     let effort = budget_tokens
                         .map(|b| crate::universal::reasoning::budget_to_effort(b, max_tokens));
+                    let canonical = if tc.thinking_budget.is_some() {
+                        Some(crate::universal::ReasoningCanonical::GoogleThinkingBudget)
+                    } else {
+                        Some(crate::universal::ReasoningCanonical::GoogleIncludeThoughts)
```
knjiang: should we just import these top-elvel?

crates/lingua/src/providers/google/adapter.rs
```diff
@@ -627,10 +675,50 @@ impl ProviderAdapter for GoogleAdapter {
...
+                        });
+
+                    for (index, text) in reasoning_texts.iter().enumerate() {
+                        let mut thought_part = serde_json::Map::new();
+                        thought_part.insert("text".into(), Value::String((*text).to_string()));
+                        thought_part.insert("thought".into(), Value::Bool(true));
+                        if index == reasoning_texts.len() - 1 {
+                            if let Some(signature) = thought_reasoning_signature {
+                                thought_part.insert(
+                                    "thoughtSignature".into(),
```
knjiang: can this be typed? `thoughSignature`

knjiang (approved):

payloads/scripts/transforms/__snapshots__/transforms.test.ts.snap
CLowbrow: Yeah I don't think there was coverage for this anywhere.

PR #309, opened 2026-06-16
Make claude logging easier

CLowbrow:
Need to get the headers back

### DIFF
--- a/crates/braintrust-llm-router/src/providers/anthropic.rs
+++ b/crates/braintrust-llm-router/src/providers/anthropic.rs
@@ -4,8 +4,10 @@ use crate::auth::AuthConfig;
 use crate::catalog::ModelSpec;
 use crate::client::{build_middleware_client, ClientSettings};
 use crate::error::{Error, Result, UpstreamHttpError};
-use crate::providers::ClientHeaders;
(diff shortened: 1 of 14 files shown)


crates/braintrust-llm-router/src/providers/anthropic.rs
```diff
@@ -4,8 +4,10 @@ use crate::auth::AuthConfig;
 use crate::catalog::ModelSpec;
 use crate::client::{build_middleware_client, ClientSettings};
 use crate::error::{Error, Result, UpstreamHttpError};
-use crate::providers::ClientHeaders;
-use crate::streaming::{sse_stream, RawResponseStream};
+use crate::providers::{
+    provider_response_with_headers, ClientHeaders, ProviderResponse, ProviderStreamResponse,
```
knjiang: lets not do this. we can just intercept at the middleware layer similar to how we did https://github.com/braintrustdata/lingua/pull/310/changes

i really don't want to break the interface from Bytes.  Concretely, i don't like `ProviderResponse` and `ProviderStreamResponse`

knjiang (changes requested):

PR #320, opened 2026-06-24
Update tool handling conversion

CLowbrow:
- Project Lingua tool discovery messages into valid Chat Completions-shaped messages.
- Convert `tool_discovery_call` to assistant `tool_calls[]`.
- Convert `tool_discovery_result` to `role: "tool"` messages with JSON string content and the matching `tool_call_id`.
- Update WASM and Python `lingua_to_chat_completions_messages` exports to use the specialized vector conversion path so grouped tool results expand correctly.
- Add an import fixture for Responses `tool_search_call` / `tool_search_output` where both items have `call_id: null`, preserving the prior call id for the result.


colsondonohue:
@knjiang can you verify this rust? idk any of the language idioms or anything

knjiang (approved):

PR #337, opened 2026-07-09
Preserve Anthropic system-prompt cache_control on OpenAI→Anthropic path

paultancre-bt:
## Problem

On the OpenAI-compat `/chat/completions` path to a Claude model, `cache_control` on the **system** message was silently dropped, so Anthropic prompt caching on the system prompt never engaged (`cached_tokens: 0`). User-message and native `/anthropic/v1/messages` paths were unaffected.

## Approach

`request_from_universal` (anthropic adapter) flattened leading system messages into a plain `String`, keeping only `t.text`. The top-level `system` param supports an array of text blocks with `cache_control`, so:

* If any system text block carries `cache_control`, emit `system` as an array of `{type,text,cache_control}` blocks (reusing `anthropic_cache_control_from_universal`, now `pub(crate)`).
* Otherwise keep the existing string flattening — no behavior change when caching isn't requested.

## Tests

* `test_anthropic_system_cache_control_preserved` — asserts outbound `system` is an array whose block carries `cache_control.type = "ephemeral"`. Verified **red without the fix, green with it**.
* `test_anthropic_system_without_cache_control_stays_string` — backward-compat guard.
* `cargo fmt --check` + `cargo clippy -p lingua` clean.

## Repro (prod, `x-bt-use-cache: never`)

Warm cache via native (`cache_creation=5012`), then read identical prefix via openai-compat system → MISS before fix. Scripts: `scripts/repro_pylon_18806_*.py` (monorepo).

Linear: [BT-6090](https://linear.app/braintrustdata/issue/BT-6090/gateway-drops-anthropic-system-prompt-cache-control-on-openaianthropic) ([BT-6090](https://linear.app/braintrustdata/issue/BT-6090))<br>Pylon: [#18806](<https://github.com/braintrustdata/lingua/issues/18806>)


knjiang:
do you mind adding a test case in `params.ts` with a system message including cache_control and then running make capture

paultancre-bt:
@knjiang only supplied an Anthropic key for the test seems this make test-typescript failing do I need to supply all the keys? (- OPENAI_API_KEY, ANTHROPIC_API_KEY, GOOGLE_API_KEY,AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY) ?

knjiang:
yep all those keys would be nice. this is to make sure we understand our change in the context of all the different model apis

erin2722 (approved):

PR #350, opened 2026-07-13
Follow up fixes for reasoning params

ankrgyl:


ankrgyl:
Once https://github.com/braintrustdata/braintrust/pull/17137 lands we should auto generate the streaming types

PR #375, opened 2026-07-22
Auto fix type generation PRs

CLowbrow:
## Summary

This PR automates follow-up fixes when Codex reviews provider type update PRs and finds actionable issues. It securely verifies that both the review and PR came from trusted automation, then uses Claude to propose a bounded fix and pushes it only after focused validation succeeds. The complete generation, review, repair, validation, and publication lifecycle is recorded under the original Braintrust trace.

- Runs only for verified Codex reviews on PRs created by the provider type update workflow.
- Limits autofix to two attempts per PR.
- Runs Claude without repository write access.
- Validates patch scope, tests, Clippy, and typed boundaries before pushing.
- Rejects stale heads, unsafe paths, oversized patches, and failed validation.
- Records the full flow under the original Braintrust trace.
- Keeps validation and publishing isolated from model credentials.
- Hardens the existing feedback workflow to execute trusted code from `main`.

### DIFF
--- a/.github/scripts/braintrust-provider-types.mjs
+++ b/.github/scripts/braintrust-provider-types.mjs
@@ -464,6 +464,9 @@ function evaluateCodexAutofixEligibility({
   return {
     eligible: true,
     provider: metadata.provider,
+    braintrustProject: metadata.project,
+    rootSpanId: metadata.root_span_id,
+    sourceSpanId: metadata.span_id,
     prNumber: pullRequest.number,
     prUrl: pullRequest.html_url,
     headRef: pullRequest.head.ref,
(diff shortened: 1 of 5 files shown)


.github/scripts/braintrust-provider-types.mjs
```diff
@@ -1,8 +1,25 @@
...
+  login: "github-actions[bot]",
+  id: 41898282,
+});
+const PROVIDER_TYPE_WORKFLOW_PATH =
+  ".github/workflows/update-provider-types.yml";
+const PROVIDER_TYPE_WORKFLOW_NAME = "Update provider types";
+const AUTOFIX_MARKER = "provider-type-codex-autofix";
+const AUTOFIX_MAX_ATTEMPTS = 2;
+const AUTOFIX_MAX_FILES = 8;
+const AUTOFIX_MAX_CHANGED_LINES = 800;
```
cpinn: wonder if this might need to be higher for the generated types

cpinn (approved):

.github/scripts/braintrust-provider-types.mjs
CLowbrow: ohhh yeah good call we get so much garbage generated stuff

PR #412, opened 2026-08-04
Tolerate out-of-enum values on pass-through params in OpenAI request detection (GATE-6)

evanmkeith:
## Problem

`POST /v1/proxy/chat/completions` with `service_tier: "fast"` returns a gateway 400 (`Unable to detect request source format`) even though the identical body succeeds against OpenAI directly. Reported in Pylon #19926 (Tavus); tracked as GATE-6.

Format detection is implemented as a strict full-body deserialization into the generated `CreateChatCompletionRequestClass` / `CreateResponseClass`, so a single out-of-enum value on an unrelated pass-through parameter makes the gateway conclude it cannot identify the request format at all. The parsed struct is discarded at every call site (`.is_ok()`), so this strictness bought nothing — and the conversion path already handles these fields permissively (`OpenAIChatParams.service_tier` is `Option<String>`, forwarded verbatim).

This is not specific to `service_tier`: every closed unit-variant enum on a top-level request param is a latent 400 (`verbosity`, `prompt_cache_retention`, `modalities`, `reasoning_effort`, `truncation`, `include` — all confirmed live against prod). It recurs every time OpenAI ships a value newer than our vendored spec snapshot.

## Fix

Add a structural fallback to detection, tried only after the strict parse fails:

- `detect_openai_shape` — requires a `model` string and a non-empty, well-formed `messages` array (the exact fields `request_to_universal` consumes). No gating on enum-valued top-level params.
- `detect_responses_shape` — preserves the `input`-present / `messages`-absent discriminator verbatim, then requires `input` to parse as `InputParam`.

Wired into `OpenAIAdapter::detect_request` / `detect_passthrough_request` (strict → legacy-prompt → shape) and `ResponsesAdapter::detect_request`. The strict parsers are untouched — `validation/openai.rs` keeps its contract, and the strict attempt still runs first so the common case is behaviorally identical. Same shape as the BT-5671 fix (`try_parse_openai_legacy_prompt`): make detection more permissive, don't loosen the generated types.

## Safety

- `OpenAIAdapter` is registered last in `ADAPTERS`, so its widened detection can only claim payloads nothing else matched.
- `ResponsesAdapter` is registered *first*, so leniency there is argued separately: the discriminator is unchanged, and since the strict parse already ignores unknown fields, the only new payloads admitted are ones with known fields carrying out-of-enum values — no new payload *shape* becomes claimable.
- The passthrough override still rejects legacy prompt-only bodies (the fallback requires `messages`), which is what that override exists to keep out of zero-copy passthrough. Response and stream detection remain strict.
- Error-shift risk: some malformed bodies that previously got `UnableToDetectRequestFormat` will now fail deeper with `ToUniversalFailed`. Both are 400s; the message is arguably more accurate.

## Known residuals (intentionally out of scope)

- `reasoning_effort` is only half-fixed: `OpenAIChatParams.reasoning_effort` is also a closed enum, so a new effort level now fails in conversion instead of detection. Widening it touches `build_reasoning_config` and cross-provider budget math.
- Closed enums *inside* messages/input items (role, content-part types) still gate detection. This PR decouples top-level params only; noted in the doc comments.
- Anthropic detection has the same strict-parse shape (`ServiceTierEnum`); same fix pattern applies if it surfaces.

## Testing

- New guard tests loop every known closed-enum top-level param on both adapters, asserting detection tolerates unknown values — and that the strict parse still rejects them, so a future spec refresh that absorbs a value flags the test case instead of silently passing. This is the test that would have caught GATE-6.
- Strict-contract pin: `try_parse_openai` still rejects `service_tier: "fast"` and accepts `"flex"`.
- Full suite: 1224 passed. Payload snapshots: 1745 passed, zero changes (detection claims no payload it previously declined). Fuzz regression suite passes. Clippy clean.
- Post-deploy: rerun the Pylon #19926 repro harness — `service_tier: "fast"` should return 200, and junk values should surface an OpenAI-originated error rather than a gateway 400.

### DIFF
--- a/crates/lingua/src/providers/openai/adapter.rs
+++ b/crates/lingua/src/providers/openai/adapter.rs
@@ -28,7 +28,9 @@ use crate::providers::openai::params::{
 };
 use crate::providers::openai::responses_adapter::parse_responses_extras;
 use crate::providers::openai::tool_parsing::parse_openai_chat_tools_array;
-use crate::providers::openai::{try_parse_openai, try_parse_openai_legacy_prompt};
+use crate::providers::openai::{
+    detect_openai_shape, try_parse_openai, try_parse_openai_legacy_prompt,
+};
 use crate::serde_json::{self, Map, Value};
 use crate::universal::convert::TryFromLLM;
 use crate::universal::message::{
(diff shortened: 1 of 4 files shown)


erin2722:
Rather than tolerating the out-of-enum values here, we will instead update the enums in https://github.com/braintrustdata/lingua/pull/410. I'll push that change forward!

PR #433, opened 2026-08-18
Always include required anthropic field

CLowbrow:
## Typespecs say this is the correct behavior:

Every non-streaming response serialized to Anthropic format includes `"stop_sequence": null` when Lingua has no matched custom stop sequence to preserve, including responses with `"stop_reason": "tool_use"`.

Docs: https://platform.claude.com/docs/en/api/typescript/messages


erin2722 (approved):

CLowbrow:
@codex review

PR #438, opened 2026-08-27
Pylon 20661 fix

stretpjc:
## Context

OpenAI Responses traces commonly contain a `reasoning` item immediately followed by a `function_call`.

Lingua correctly imports this as one assistant message containing both reasoning and a tool call. When converted to Chat Completions format for replay, Lingua also emits `reasoning_signature` as an array to preserve signature provenance.

That array is necessary for lossless replay, but it is not accepted by Braintrust’s display schema. Display parsing can then fall through to another message parser that silently removes the OpenAI-specific `tool_calls` field.

## Description

This PR adds a typed, display-specific Chat Completions conversion owned by Lingua.

- Add `messages_to_chat_completion_display_messages` in the OpenAI converter.
- Expose it through WASM and TypeScript as `linguaToChatCompletionsDisplayMessages`.
- Preserve reasoning text and top-level tool calls in display output.
- Omit encrypted `reasoning_signature` values from display output.
- Leave the existing `linguaToChatCompletionsMessages` replay conversion unchanged and lossless.
- Add an import fixture for `reasoning` immediately followed by `function_call`.
- Add Rust and TypeScript regression coverage for the new conversion and exports.

No generated provider types or provider request/response transforms are changed.

## Testing

Tested locally following these steps:

- Built Braintrust locally with the Lingua submodule checked out at [`2a175fbe`](https://github.com/braintrustdata/lingua/commit/2a175fbea44a7e666a96f7fa0aa69b879ca00276).
- Applied the web-app integration changes from Braintrust commit [`14d8281b`](https://github.com/braintrustdata/braintrust/commit/14d8281bc96dca268b0c66855d01c800051fe735) on the [`pylon-20661-move-to-lingua`](https://github.com/braintrustdata/braintrust/tree/pylon-20661-move-to-lingua) branch.
- Loaded the synthetic reproduction traces and confirmed that tool calls render when a `function_call` is immediately preceded by a `reasoning` item.

This the same trace in staging (left) vs local (right):
Screenshot 2026-08-27 at 12 59 04 PM


stretpjc:
Addressed the Codex feedback by moving replay-only signature cleanup into Lingua’s display conversion path, stripping those signatures before validation while preserving reasoning content and tool calls, and adding coverage across Rust, import fixtures, WASM, and TypeScript exports to verify the behavior end to end.

Tested locally with the same steps as before but pointing to the updated lingua commit. The tests had the same behavior. Not marking Codex's comments resolved just for reference

knjiang (approved):

PR #442, opened 2026-08-31
Fix crypto lib resolution for vertex service account key

erin2722:
see https://github.com/braintrustdata/braintrust/pull/19686 for context


Cargo.toml
```diff
@@ -97,7 +97,8 @@ aws-smithy-eventstream = { version = "0.60.20" }
 aws-smithy-types = { version = "1.4.7" }
 
 # Auth
-jsonwebtoken = { version = "10.3", features = ["rust_crypto"] }
+aws-lc-rs = "=1.15.0"
+jsonwebtoken = { version = "10.3", default-features = false, features = ["aws_lc_rs", "use_pem"] }
```
erin2722, line 101: see the main repo explanation, but I figured it made sense to centralize on `aws_lc_rs` instead of having multiple crypto providers around, this prunes our deps tree quite a bit.

crates/braintrust-llm-router/src/auth/google.rs
```diff
@@ -13,6 +13,17 @@ const DEFAULT_TOKEN_URI: &str = "https://oauth2.googleapis.com/token";
...
+    let provider = &jsonwebtoken::crypto::aws_lc::DEFAULT_PROVIDER;
+    match provider.install_default() {
+        Ok(()) => Ok(()),
+        Err(existing) if std::ptr::eq(existing, provider) => Ok(()),
+        Err(_) => Err(anyhow!(
+            "a non-AWS-LC jsonwebtoken provider was already installed"
+        )),
+    }
+}
+
```
knjiang, line 26: how dangerous is this? or perf-wise is thsi okay?  like is this a one time cost thing?

knjiang (approved):

PR #458, opened 2026-09-17
fix: add evaluation flavor to ModelFlavor enum

erin2722:
## Summary

The shared model catalog (`proxy` `model_list.json`) now includes TypeSafe `jev-*` models with `"flavor": "evaluation"`, added in braintrustdata/braintrust-proxy#1267. The Rust `ModelFlavor` enum only knew `chat` / `completion` / `embedding` / `responses`, so deserializing the catalog fails with `unknown variant "evaluation"`, panicking gateway/api-rs startup.

This adds the `Evaluation` variant so the shared catalog deserializes.

## Notes

- `#[serde(rename_all = "lowercase")]` maps the variant to `"evaluation"`.
- No exhaustive `match` on `ModelFlavor` exists in the tree, so this is purely additive.

## Validation

- `cargo test -p braintrust-llm-router` passes.
- In the consuming `braintrust` repo (with the proxy submodule bumped to include #1267): the previously-panicking gateway tests now pass — `cargo test -p gateway --lib proxy::router_builder::tests` (36 passed) and `server::tests::supports_modern_and_legacy_prefixes` — and `cargo fmt -- --check` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


erin2722:
I think we can just add the type now and then fill in the actual lingua changes once we are ready

knjiang (approved):

